PR #14, opened 2016-12-15
Import wizard

lmsilvera:
https://www.pivotaltracker.com/story/show/135269673

### DIFF
--- a/app/Resources/assets/js/lsdoc/index.js
+++ b/app/Resources/assets/js/lsdoc/index.js
@@ -1,16 +1,29 @@
 $(document).on('ready', function(){
-    SaltGithub.getRepoList();
+    SaltGithub.init();
+    $('input[name="import"]').click(function(){
+        loadContent($('input[name="import"]:checked').val());
+    });
 });
 
 var SaltGithub = (function(){
 
+    var githubData = null;
+
+    function init(){
+        getRepoList();
+    }
+
     function getRepoList(){
         if( $('.js-github-list').length > 0 ){
-            $.get('/app_dev.php/user/github/files', function(data){
-                $.each(data.data, function(i, e){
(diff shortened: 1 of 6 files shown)


app/Resources/assets/js/lsdoc/index.js
```diff
@@ -1,16 +1,33 @@
...
+        getRepoList();
+        $('#js-wizard-btn').click(function(){
+            loadModal();
+        });
+    }
+
     function getRepoList(){
         if( $('.js-github-list').length > 0 ){
-            $.get('/app_dev.php/user/github/files', function(data){
+            $.get('/app_dev.php/user/github/repos', function(data){
```
roverwolf: We should only use /app_dev.php for development, as it will not work for normal users at some point after we have deployed to a production system.

src/CftfBundle/Resources/views/LsDoc/index.html.twig
```diff
@@ -1,58 +1,102 @@
...
+            <div class="modal-header">
+                <h4>Import Framework</h4>
+            </div>
+            <div class="modal-body">
+                <div class="row">
+                    <div class="col-md-12">
+                        ASN:
+                        <input name="import" type="radio" value="asn" checked="checked"></input>
+                        Github:
+                        <input name="import" type="radio" value="github"></input>
```
roverwolf: This should probably not show up unless github has been configured (github_client_id has a useful value), otherwise is guaranteed to just end up erroring for the end user.

We should probably also look at adding some documentation for an administrator on how to get the github client id and secret and how to configure it (part of a more general question about documentation).

roverwolf:
I went ahead and made the changes that I suggested, and then merged the branch into develop.

PR #24, opened 2017-01-18
[WIP] Import from GitHub json

roverwolf:
This change is [Reviewable](https://reviewable.io/reviews/opensalt/opensalt/24)


src/CftfBundle/Resources/views/LsDoc/index.html.twig
```diff
@@ -43,10 +43,8 @@
                     <div class="col-md-12">
                         ASN:
                         <input name="import" type="radio" value="asn" checked="checked"></input>
-                        {% if github_client_id is not empty %}
```
roverwolf: Was this removed on purpose?

If the specific instance of Salt does not have a github_client_id set in the parameters file (which also requires it to be registered with GitHub) then there is no way any of the github importing can work.

src/GithubFilesBundle/Controller/DefaultController.php
```diff
@@ -19,18 +19,22 @@ class DefaultController extends Controller
     /**
      * @Route("/user/github/repos")
      */
-    public function reposAction()
+    public function getRepos(Request $request)
```
roverwolf: route actions should end with "Action", so this should probably be renamed to getReposAction

src/GithubFilesBundle/Service/GithubImport.php
```diff
@@ -0,0 +1,166 @@
...
+                $lsDoc->setOfficialUri($this->getValue($lsDocKeys['officialSourceURL'], $row));
+                $lsDoc->setPublisher($this->getValue($lsDocKeys['publisher'], $row));
+                $lsDoc->setDescription($this->getValue($lsDocKeys['description'], $row));
+                $lsDoc->setVersion($this->getValue($lsDocKeys['version'], $row));
+                $lsDoc->setSubject($this->getValue($lsDocKeys['subject'], $row));
+                $lsDoc->setLanguage($this->getValue($lsDocKeys['language'], $row));
+                $lsDoc->setNote($this->getValue($lsDocKeys['notes'], $row));
+
+                $em->persist($lsDoc);
+                $em->flush();
```
roverwolf: This $em->flush() line should be moved to after the foreach() is complete.

src/GithubFilesBundle/Service/GithubImport.php
```diff
@@ -0,0 +1,166 @@
...
+        $em->persist($lsItem);
+
+        $lsAssociation = new LsAssociation();
+        $lsAssociation->setLsDoc($lsDoc);
+        $lsAssociation->setOrigin($lsItem);
+        $lsAssociation->setType(LsAssociation::EXACT_MATCH_OF);
+        $lsAssociation->setDestinationNodeIdentifier($lsItemIdentifier);
+
+        $em->persist($lsAssociation);
+        $em->flush();
```
roverwolf: This should probably be removed, and let saveGithubDocument() do the one flush for the entire set of entities.

src/GithubFilesBundle/Service/GithubImport.php
```diff
@@ -0,0 +1,166 @@
...
+        return $this->em;
+    }
+
+    /**
+     * Parse an Github document into a LsDoc/LsItem hierarchy
+     *
+     * @param array $lsdocKeys
+     * @param string $fileContent
+     */
+    public function parseGithubDocument($lsDocKeys, $lsItemKeys, $fileContent){
```
roverwolf: Should parseGithubDocument (and saveGithubDocument, ...) be renamed to be specific to the import of CSV files?

Or possibly renaming the class/service if different document types than csv would be imported from GitHub via a different service.

src/GithubFilesBundle/Service/GithubImport.php
```diff
@@ -0,0 +1,166 @@
...
+     *
+     * @param array $lsdocKeys
+     * @param string $content
+     */
+    public function saveGithubDocument($lsDocKeys, $lsItemKeys, $content){
+        $em = $this->getEntityManager();
+        $lsDoc = new LsDoc();
+
+        foreach($content as $i => $row){
+            if($i < 1){
```
roverwolf: The section in this if block should probably go before the foreach to avoid having to do the if check for every line since it is always done once per loaded file.

src/GithubFilesBundle/Service/GithubImport.php
```diff
@@ -0,0 +1,166 @@
...
+                //create the new LsDoc
+                $lsDocIdentifier = Uuid::uuid5(Uuid::NAMESPACE_URL, $this->getValue($lsDocKeys['title'], $row))->toString();
```
roverwolf: NAMESPACE_URL is probably not the best uuid namespace to use, since it is meant for URLs.

I am assuming from this that $lsDocKeys['title'] would never be duplicated for a different imported document... otherwise this will generate a duplicate key.

src/GithubFilesBundle/Service/GithubImport.php
```diff
@@ -0,0 +1,166 @@
...
+    }
+
+    public function parseGithubStandard(LsDoc $lsDoc, $lsItemKeys, $data){
+        $lsItem = new LsItem();
+        $em = $this->getEntityManager();
+
+        $lsItem->setLsDoc($lsDoc);
+        $lsItem->setFullStatement($this->getValue($lsItemKeys['fullStatement'], $data));
+
+        $lsItemIdentifier = Uuid::uuid5(Uuid::NAMESPACE_URL, $this->getValue($lsItemKeys['fullStatement'], $data))->toString();
```
roverwolf: NAMESPACE_URL is probably not the best uuid namespace to use, since it is meant for URLs.

I am pretty certain that $lsDocKeys['fullStatement'] is not guaranteed to be unique.  Are we sure that we want to be generating the same UUID as the identifier each time the same statement appears (even in different imported files)?

app/Resources/assets/js/lsdoc/index.js
```diff
@@ -1,32 +1,44 @@
...
         $('#js-wizard-btn').click(function(){
             loadModal();
         });
     }
 
-    function getRepoList(){
+    function getRepoList(page, perPage){
         if( $('.js-github-list').length > 0 ){
-            $.get('/user/github/repos', function(data){
+            $.get('/app_dev.php/user/github/repos', { page: page, perPage: perPage }, function(data){
```
roverwolf: We will want to ensure that all /app_dev.php/ URLs are removed once the code is ready.

We might want to use a javascript variable here that is set on the page where it is used so that it can automatically use either the dev or non-dev path as appropriate.

app/Resources/assets/js/lsdoc/index.js
```diff
@@ -131,5 +168,217 @@ var ImportFrameworks = (function(){
...
+
+    function jsonImporter(file){
+        var json = JSON.parse(file);
+        var keys = Object.keys(json);
+        var subKey = [];
+        console.log(keys);
+
+        keys.forEach(function(subJson){
+            subKey = Object.keys(subJson);
+            console.log(subKey);
```
roverwolf: We will want to make sure that we remove console.log() calls before finishing the PR as well, as some browsers do not have that method and will then throw an exception.

src/GithubFilesBundle/Service/GithubImport.php
```diff
@@ -0,0 +1,166 @@
...
+                $lsDoc->setSubject($this->getValue($lsDocKeys['subject'], $row));
+                $lsDoc->setLanguage($this->getValue($lsDocKeys['language'], $row));
+                $lsDoc->setNote($this->getValue($lsDocKeys['notes'], $row));
+
+                $em->persist($lsDoc);
+                $em->flush();
+                continue;
+            }
+
+            $lsItem = $this->parseGithubStandard($lsDoc, $lsItemKeys, $row);
```
roverwolf: `$lsItem` will always be null here as `parseGithubStandard` does not return a value.

src/GithubFilesBundle/Service/GithubImport.php
```diff
@@ -81,53 +81,46 @@ public function parseGithubDocument($lsDocKeys, $lsItemKeys, $fileContent){
...
-                $lsDoc->setSubject($this->getValue($lsDocKeys['subject'], $row));
-                $lsDoc->setLanguage($this->getValue($lsDocKeys['language'], $row));
-                $lsDoc->setNote($this->getValue($lsDocKeys['notes'], $row));
-
-                $em->persist($lsDoc);
-                $em->flush();
-                continue;
-            }
+        //create the new LsDoc
+        $lsDoc->setUri('local:'.$lsDoc->getIdentifier());
```
roverwolf: If you don't change the identifier then you don't need to set the Uri either.

src/GithubFilesBundle/Service/GithubImport.php
```diff
@@ -81,53 +81,46 @@ public function parseGithubDocument($lsDocKeys, $lsItemKeys, $fileContent){
...
         $lsItem = new LsItem();
         $em = $this->getEntityManager();
 
         $lsItem->setLsDoc($lsDoc);
         $lsItem->setFullStatement($this->getValue($lsItemKeys['fullStatement'], $data));
 
-        $lsItemIdentifier = Uuid::uuid5(Uuid::NAMESPACE_URL, $this->getValue($lsItemKeys['fullStatement'], $data))->toString();
-        $lsItem->setIdentifier($lsItemIdentifier);
-        $lsItem->setUri('local:'.$lsItemIdentifier);
+        $lsItem->setUri('local:'.$lsItem->getIdentifier());
```
roverwolf: Same here, the Uri does not need to be set since the identifier was not changed from what __construct() did.

src/GithubFilesBundle/Service/GithubImport.php
```diff
@@ -144,10 +137,11 @@ public function parseGithubStandard(LsDoc $lsDoc, $lsItemKeys, $data){
         $lsAssociation->setLsDoc($lsDoc);
         $lsAssociation->setOrigin($lsItem);
         $lsAssociation->setType(LsAssociation::EXACT_MATCH_OF);
-        $lsAssociation->setDestinationNodeIdentifier($lsItemIdentifier);
+        $lsAssociation->setDestinationNodeIdentifier($lsItem->getIdentifier());
```
roverwolf: I'm not sure creating this association makes sense.

Basically it is saying that the lsItem is a match to itself and does not add any other information.

src/GithubFilesBundle/Service/GithubImport.php
```diff
@@ -81,53 +81,46 @@ public function parseGithubDocument($lsDocKeys, $lsItemKeys, $fileContent){
...
+        $lsDoc->setVersion($this->getValue($lsDocKeys['version'], $content[0]));
+        $lsDoc->setSubject($this->getValue($lsDocKeys['subject'], $content[0]));
+        $lsDoc->setLanguage($this->getValue($lsDocKeys['language'], $content[0]));
+        $lsDoc->setNote($this->getValue($lsDocKeys['notes'], $content[0]));
+
+        $em->persist($lsDoc);
+
+        for($i=1; $i < sizeof($content); $i++){
+            $row = $content[$i];
+            $lsItem = $this->parseCSVGithubStandard($lsDoc, $lsItemKeys, $row);
```
roverwolf: `$lsItem` is never used, so we can probably not do the assignment here....

But, we may want to add them to an array if we look at doing another pass to add any parent/child associations from the data.
Or we could just do an explicit `$lsDoc->addTopLsItem($lsItem);` to have the document be the parent of all the statements (in general the items should be part of a parent/child hierarchy).

src/GithubFilesBundle/Controller/DefaultController.php
```diff
@@ -46,16 +50,33 @@ public function getFilesAction(Request $request){
...
+            'sha' => $sha
         ]);
 
         return $response->setData(array(
-            'data' => $api->decode($listFiles)
+            'data' => $api->decode($blob)
         ));
     }
+
+    public static function parseLink($link, $rel){
```
roverwolf: Why is this a static function?  Should it be moved to a Util class instead of the controller?

src/GithubFilesBundle/Controller/DefaultController.php
```diff
@@ -19,18 +19,22 @@ class DefaultController extends Controller
...
      */
-    public function reposAction()
+    public function getReposAction(Request $request)
     {
+        $page = $request->query->get('page');
+        $perPage = $request->query->get('perPage');
+
         $currentUser = $this->getUser();
         $response = new JsonResponse();
         $token = new \Milo\Github\OAuth\Token($currentUser->getGithubToken());
```
roverwolf: We should probably check that the user has a token (at least isn't empty) before trying to use it.

If the user does not have a token then we can give a better message to the user than "an error occurred or..." and just say that they need to log in.

Also, please document the procedure on registering a client with GitHub and what the correct value of the github_redirect_uri parameter should be.

roverwolf:
Reviewed 4 of 11 files at r1, 1 of 3 files at r2, 2 of 4 files at r3, 1 of 1 files at r4.
Review status: 8 of 10 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/opensalt/opensalt/24)*

roverwolf:
Reviewed 1 of 11 files at r1, 1 of 4 files at r3.
Review status: all files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---

*[app/Resources/assets/js/lsdoc/index.js, line 31 at r5](https://reviewable.io:443/reviews/opensalt/opensalt/24#-KayK6BMRUS6xYdhWsEj:-KayK6BMRUS6xYdhWsEk:b984hip) ([raw file](https://github.com/opensalt/opensalt/blob/225568db9cdc2aa6d165b3ebb75a8e46c7fbbad3/app/Resources/assets/js/lsdoc/index.js#L31)):*
> ```JavaScript
>     function getRepoList(page, perPage){
>         if( $('.js-github-list').length > 0 ){
>             $.get('/app_dev.php/user/github/repos', { page: page, perPage: perPage }, function(data){
> ```

We shouldn't commit URLs with `app_dev.php`.

---

*[app/Resources/assets/js/lsdoc/index.js, line 55 at r5](https://reviewable.io:443/reviews/opensalt/opensalt/24#-KayKkEW06KlHDSG3a1M:-KayKkEW06KlHDSG3a1N:b640f42) ([raw file](https://github.com/opensalt/opensalt/blob/225568db9cdc2aa6d165b3ebb75a8e46c7fbbad3/app/Resources/assets/js/lsdoc/index.js#L55)):*
> ```JavaScript
> 
>         $.ajax({
>             url: '/app_dev.php/user/github/files',
> ```

Another `app_dev.php` URL.

---

*[app/Resources/assets/js/lsdoc/index.js, line 228 at r5](https://reviewable.io:443/reviews/opensalt/opensalt/24#-KayKn1k8QKcrvLCTG6T:-KayKn1l3Dt8SoidLgOo:b-6q0969) ([raw file](https://github.com/opensalt/opensalt/blob/225568db9cdc2aa6d165b3ebb75a8e46c7fbbad3/app/Resources/assets/js/lsdoc/index.js#L228)):*
> ```JavaScript
> 
>         $.ajax({
>             url: '/app_dev.php/cf/github/import',
> ```

One more `app_dev.php` URL that we should fix.

---

*[src/CftfBundle/Resources/views/LsDoc/index.html.twig, line 46 at r5](https://reviewable.io:443/reviews/opensalt/opensalt/24#-KayKzEjrzTXJtFUQr0S:-KayKzEjrzTXJtFUQr0T:b9gfadi) ([raw file](https://github.com/opensalt/opensalt/blob/8da1bcfc976087762f5f24e965125bc0b095de8c/src/CftfBundle/Resources/views/LsDoc/index.html.twig#L46)):*
> ```Twig
>                         ASN:
>                         <input name="import" type="radio" value="asn" checked="checked"></input>
>                         {% if github_client_id is not empty %}
> ```

We should probably keep the check to see if the application is configured as registered with GitHub (has the `github_client_id` parameter set).

---

*[src/CftfBundle/Resources/views/LsDoc/index.html.twig, line 63 at r5](https://reviewable.io:443/reviews/opensalt/opensalt/24#-KayLM2Tk53RLFUe3UT-:-KayLM2Tk53RLFUe3UT0:b-iog14a) ([raw file](https://github.com/opensalt/opensalt/blob/225568db9cdc2aa6d165b3ebb75a8e46c7fbbad3/src/CftfBundle/Resources/views/LsDoc/index.html.twig#L63)):*
> ```Twig
>                                         <div id="files" class="hidden"></div>
>                                         <p class="js-github-message-error" style="display: none;">
>                                             {% if github_client_id is not empty %}
> ```

If we do the check for `github_client_id` above, then it is probably not needed here.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/opensalt/opensalt/24)*

roverwolf:
Reviewed 2 of 3 files at r6.
Review status: 9 of 10 files reviewed at latest revision, 2 unresolved discussions, some commit checks broke.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/opensalt/opensalt/24)*

roverwolf:
Review status: 9 of 10 files reviewed at latest revision, all discussions resolved, some commit checks broke.

---

*[src/CftfBundle/Resources/views/LsDoc/index.html.twig, line 46 at r5](https://reviewable.io:443/reviews/opensalt/opensalt/24#-KayKzEjrzTXJtFUQr0S:-KbHe5Km3a5u-XKD4U27:b-e5v0qx) ([raw file](https://github.com/opensalt/opensalt/blob/8da1bcfc976087762f5f24e965125bc0b095de8c/src/CftfBundle/Resources/views/LsDoc/index.html.twig#L46)):*

We should probably keep the check to see if the application is configured as registered with GitHub (has the `github_client_id` parameter set).
</blockquote></details>

I've moved the check for github_client_id to not show the GitHub option if it does not exist.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/opensalt/opensalt/24)*

roverwolf:
Reviewed 1 of 1 files at r7.
Review status: all files reviewed at latest revision, all discussions resolved, some commit checks broke.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/opensalt/opensalt/24)*

PR #26, opened 2017-01-25
Add github auth config documentation and updating readme

lmsilvera:
This change is [Reviewable](https://reviewable.io/reviews/opensalt/opensalt/26)

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ Installation
 
 2. Create env file
   ```
-  cp docker/.env.dist to docker/.env
+  cp docker/.env.dist docker/.env
   ```
 
 3. Install components with composer
@@ -69,3 +69,4 @@ Other Docs
 ----------
 
 - [User Management Commands](./src/Salt/UserBundle/Resources/doc/Commands.md)
+- [Github Authentication Config](./src/GithubFilesBundle/Resources/doc/GithubAuth.md)
--- a/src/GithubFilesBundle/Resources/doc/GithubAuth.md
+++ b/src/GithubFilesBundle/Resources/doc/GithubAuth.md
@@ -8,13 +8,11 @@ You need have your Github Access Keys (`client_id` and `client_secret`). If you 
 #### 2. Setup OpenSalt
 Add github keys on `config/app/config/parameters.yml`.
 
-Replace values of keys `github_redirect_uri`, `github_client_id`, `github_client_secret`. `parameters.yml` will looks like this:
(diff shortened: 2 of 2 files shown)


roverwolf:
Review status: 0 of 3 files reviewed at latest revision, 1 unresolved discussion, some commit checks broke.

---

*[app/config/parameters.yml.dist, line 25 at r1](https://reviewable.io:443/reviews/opensalt/opensalt/26#-KbQTIomCwmNw4iPWYZp:-KbQTIomCwmNw4iPWYZq:byv1fjv) ([raw file](https://github.com/opensalt/opensalt/blob/bbfb43d729e308fa029545ba2d79158705468d1c/app/config/parameters.yml.dist#L25)):*
> ```
> 
>     local_uri_host: http://localhost:3000/
>     github_redirect_uri: ~
> ```

Is there a reason we need to pass this in instead of having the code generate the absolute path to the route that this is pointing to?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/opensalt/opensalt/26)*

lmsilvera:
@roverwolf because the component that I use for authenticate users with github send the github_redirect_uri with null value and then github redirect with an error like "uri mistmach" cause by the redirect_uri should be the equals.

roverwolf:
I understand that we need to send the value, but I think it should be able
to be generated based on the route to the salt endpoint that GitHub would
use.

I am assuming that the 'github_login' route would be the correct one, so to
generate the URL to pass would use something like

```
$this->generateUrl('github_login', [], UrlGeneratorInterface::ABSOLUTE_URL)
```

lmsilvera:
Hm, make sense.

lmsilvera:
@roverwolf PR updated & ready.

roverwolf:
Reviewed 1 of 3 files at r1, 1 of 1 files at r2, 1 of 1 files at r4.
Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[a discussion](https://reviewable.io:443/reviews/opensalt/opensalt/26#-KbkkFAn0Spx_BgwbVDH:-KbkkFAn0Spx_BgwbVDI:bsd1xp0) (no related file):*
@lmsilvera I've added information to the readme file to show explain the two methods of setting the github access keys.

If you agree with the way it is worded then I'll merge the PR.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/opensalt/opensalt/26)*

PR #90, opened 2017-07-24
Implementing jquery-comments to handle the commenting system

JorgeLu1s:
this is the backend and frontend for the commenting system, I'm not sure about some code in the backend, so any feedback is welcome

---
This change is [Reviewable](https://reviewable.io/reviews/opensalt/opensalt/90)


roverwolf:
Reviewed 12 of 12 files at r1.
Review status: all files reviewed at latest revision, 19 unresolved discussions, some commit checks failed.

---

*[app/DoctrineMigrations/Version20170719163804.php, line 35 at r1](https://reviewable.io:443/reviews/opensalt/opensalt/90#-Kptx5krvGVQzlj1Im2g:-Kptx5krvGVQzlj1Im2h:b-kjpavc) ([raw file](https://github.com/opensalt/opensalt/blob/db6c0c100a8509155a26f5db1d43c9802384d2af/app/DoctrineMigrations/Version20170719163804.php#L35)):*
<details><summary><i alt="summary">Quoted 14 lines of code…</i></summary>
>        $this->addSql('ALTER TABLE ls_def_subject CHANGE updated_at updated_at DATETIME DEFAULT CURRENT_TIMESTAMP NOT NULL');
>         $this->addSql('ALTER TABLE ls_def_item_type CHANGE updated_at updated_at DATETIME DEFAULT CURRENT_TIMESTAMP NOT NULL');
>         $this->addSql('ALTER TABLE ls_item DROP type, CHANGE updated_at updated_at DATETIME DEFAULT CURRENT_TIMESTAMP NOT NULL');
>         $this->addSql('ALTER TABLE rubric_criterion CHANGE updated_at updated_at DATETIME DEFAULT CURRENT_TIMESTAMP NOT NULL');
>         $this->addSql('ALTER TABLE ls_association CHANGE updated_at updated_at DATETIME DEFAULT CURRENT_TIMESTAMP NOT NULL');
>         $this->addSql('ALTER TABLE ls_def_licence CHANGE updated_at updated_at DATETIME DEFAULT CURRENT_TIMESTAMP NOT NULL');
>         $this->addSql('ALTER TABLE ls_def_association_grouping CHANGE updated_at updated_at DATETIME DEFAULT CURRENT_TIMESTAMP NOT NULL');
>         $this->addSql('ALTER TABLE import_logs CHANGE is_read is_read TINYINT(1) NOT NULL');
>         $this->addSql('ALTER TABLE ls_def_grade CHANGE updated_at updated_at DATETIME DEFAULT CURRENT_TIMESTAMP NOT NULL');
>         $this->addSql('ALTER TABLE rubric_criterion_level CHANGE updated_at updated_at DATETIME DEFAULT CURRENT_TIMESTAMP NOT NULL');
>         $this->addSql('ALTER TABLE ls_doc CHANGE updated_at updated_at DATETIME DEFAULT CURRENT_TIMESTAMP NOT NULL');
>         $this->addSql('ALTER TABLE rubric CHANGE updated_at updated_at DATETIME DEFAULT CURRENT_TIMESTAMP NOT NULL');
>         $this->addSql('ALTER TABLE ls_def_concept CHANGE updated_at updated_at DATETIME DEFAULT CURRENT_TIMESTAMP NOT NULL');
>         $this->addSql('ALTER TABLE salt_user CHANGE locked locked TINYINT(1) NOT NULL');
</details>

Any lines that are the same in both `up` and `down` should be removed.  The auto-generation of migrations does not do a good job when there is custom column definition (such as for `updated_at`).

---

*[app/DoctrineMigrations/Version20170724205513.php, line 35 at r1](https://reviewable.io:443/reviews/opensalt/opensalt/90#-KptxlSYfFMeYJ9cBYs3:-KptxlSYfFMeYJ9cBYs4:b-c2clsn) ([raw file](https://github.com/opensalt/opensalt/blob/db6c0c100a8509155a26f5db1d43c9802384d2af/app/DoctrineMigrations/Version20170724205513.php#L35)):*
> ```PHP
>         $this->addSql('ALTER TABLE comments ADD item VARCHAR(255) NOT NULL, DROP item_id, CHANGE parent parent INT DEFAULT NULL');
>     }
> ```

Same issue here, the extra alter statements are not needed when they do not actually change anything.

We might want to merge the alter to the comments table into the previous migration as well.

---

*[app/Resources/assets/js/cftree/view-trees.js, line 268 at r1](https://reviewable.io:443/reviews/opensalt/opensalt/90#-KptyUpqp1xy-LxmA6CU:-KptyUpqp1xy-LxmA6CV:b-xh87va) ([raw file](https://github.com/opensalt/opensalt/blob/db6c0c100a8509155a26f5db1d43c9802384d2af/app/Resources/assets/js/cftree/view-trees.js#L268)):*
> ```JavaScript
> 				// launching comment system depending of the item id.
> 				CommentSystem.init(item);
> ```

Spaces should be used instead of tabs.

---

*[app/Resources/assets/js/lsdoc/index.js, line 555 at r1](https://reviewable.io:443/reviews/opensalt/opensalt/90#-Kptye9DhmsRF9vXHv_H:-Kptye9DhmsRF9vXHv_I:b-xh87va) ([raw file](https://github.com/opensalt/opensalt/blob/db6c0c100a8509155a26f5db1d43c9802384d2af/app/Resources/assets/js/lsdoc/index.js#L555)):*
> ```JavaScript
> 
>     function init(nodeRef){
> 	setItem(nodeRef);
> ```

Spaces should be used instead of tabs.

---

*[app/Resources/assets/js/lsdoc/index.js, line 613 at r1](https://reviewable.io:443/reviews/opensalt/opensalt/90#-KptyhMYldVfz--000m0:-KptyhMYldVfz--000m1:b-h8onne) ([raw file](https://github.com/opensalt/opensalt/blob/db6c0c100a8509155a26f5db1d43c9802384d2af/app/Resources/assets/js/lsdoc/index.js#L613)):*
> ```JavaScript
> 
>     return {
> 	init: init
> ```

Spaces should be used instead of tabs here as well.

---

*[app/Resources/views/base.html.twig, line 16 at r1](https://reviewable.io:443/reviews/opensalt/opensalt/90#-KptytLNWArOlGyniGQL:-KptytLNWArOlGyniGQM:bi83y77) ([raw file](https://github.com/opensalt/opensalt/blob/db6c0c100a8509155a26f5db1d43c9802384d2af/app/Resources/views/base.html.twig#L16)):*
> ```Twig
>         {% block stylesheets %}
>             <link rel="stylesheet" href="{{ asset('css/main.css'|asset_version) }}" />
>             <link rel="stylesheet" href="{{ asset('css/jquery-comments.css') }}" />
> ```

Should probably add `|asset_version` here as well to help with cache busting.

---

*[src/CftfBundle/Resources/views/DocTree/view.html.twig, line 425 at r1](https://reviewable.io:443/reviews/opensalt/opensalt/90#-KptzAIrNp5fvRMZXbCT:-KptzAIrNp5fvRMZXbCU:byi681h) ([raw file](https://github.com/opensalt/opensalt/blob/db6c0c100a8509155a26f5db1d43c9802384d2af/src/CftfBundle/Resources/views/DocTree/view.html.twig#L425)):*
> ```Twig
> 
> {% block javascripts %}
> <script src="{{ asset('js/jquery-comments.min.js') }}"></script>
> ```

This might need `|asset_version` to help with cache busting.

---

*[src/Salt/SiteBundle/Controller/CommentsController.php, line 24 at r1](https://reviewable.io:443/reviews/opensalt/opensalt/90#-Kpu5aHpXAoEHEImKeXz:-Kpu5aHpXAoEHEImKeY-:b-4cj8f5) ([raw file](https://github.com/opensalt/opensalt/blob/db6c0c100a8509155a26f5db1d43c9802384d2af/src/Salt/SiteBundle/Controller/CommentsController.php#L24)):*
> ```PHP
> 
>         $comment->setContent($request->request->get('content'));
>         $comment->setCommentId($request->request->get('id'));
> ```

Is an arbitrary value okay here?  There is no validation on what is received as `id`.

---

*[src/Salt/SiteBundle/Controller/CommentsController.php, line 26 at r1](https://reviewable.io:443/reviews/opensalt/opensalt/90#-Kpu1mtJYuTLdDcfRzWr:-Kpu1mtJYuTLdDcfRzWs:byskcyk) ([raw file](https://github.com/opensalt/opensalt/blob/db6c0c100a8509155a26f5db1d43c9802384d2af/src/Salt/SiteBundle/Controller/CommentsController.php#L26)):*
> ```PHP
>         $comment->setCommentId($request->request->get('id'));
>         $comment->setParent(
>             empty($request->request->get('parent'))?0:$request->request->get('parent')
> ```

Why a `0` instead of `null`?  The 0 gets turned into null when listing the comment below, so does 0 mean anything?

Can `parent` be a FK?  Or does the field have additional structure?

---

*[src/Salt/SiteBundle/Controller/CommentsController.php, line 30 at r1](https://reviewable.io:443/reviews/opensalt/opensalt/90#-Kpu3j7UAXRTb8MAe4Xk:-Kpu3j7UAXRTb8MAe4Xl:bq1orjc) ([raw file](https://github.com/opensalt/opensalt/blob/db6c0c100a8509155a26f5db1d43c9802384d2af/src/Salt/SiteBundle/Controller/CommentsController.php#L30)):*
> ```PHP
>         $comment->setUserId($user->getId());
>         $comment->setFullname($user->getUsername().' - '.$user->getOrg()->getName());
>         $comment->setUpvoteCount($request->request->get('upvote_count'));
> ```

Would `upvoteCount` ever not be 0 when the comment is initially created?  Is an arbitrary value okay to accept here?

---

*[src/Salt/SiteBundle/Controller/CommentsController.php, line 31 at r1](https://reviewable.io:443/reviews/opensalt/opensalt/90#-Kpu55epyq6vnI6VK6xN:-Kpu55epyq6vnI6VK6xO:bm8j9kj) ([raw file](https://github.com/opensalt/opensalt/blob/db6c0c100a8509155a26f5db1d43c9802384d2af/src/Salt/SiteBundle/Controller/CommentsController.php#L31)):*
> ```PHP
>         $comment->setFullname($user->getUsername().' - '.$user->getOrg()->getName());
>         $comment->setUpvoteCount($request->request->get('upvote_count'));
>         $comment->setUserHasUpvoted(false);
> ```

This should probably be set as the default in the entity, so wouldn't have to set explicitly here.

---

*[src/Salt/SiteBundle/Controller/CommentsController.php, line 32 at r1](https://reviewable.io:443/reviews/opensalt/opensalt/90#-KptzP2XPBoOxwPN_0vd:-KptzP2XPBoOxwPN_0ve:bdolu35) ([raw file](https://github.com/opensalt/opensalt/blob/db6c0c100a8509155a26f5db1d43c9802384d2af/src/Salt/SiteBundle/Controller/CommentsController.php#L32)):*
> ```PHP
>         $comment->setUpvoteCount($request->request->get('upvote_count'));
>         $comment->setUserHasUpvoted(false);
>         $comment->setItem($request->request->get('itemType').$request->request->get('itemId'));
> ```

We might want to separate the components of the `item` reference with something like a `:`, or possibly even use a real FK.

There is also no verification that such an itemType + itemId exists here, nor that the user has access to view it.

---

*[src/Salt/SiteBundle/Controller/CommentsController.php, line 34 at r1](https://reviewable.io:443/reviews/opensalt/opensalt/90#-Kpu2sLubXEV698QoDVd:-Kpu2sLubXEV698QoDVe:bshvova) ([raw file](https://github.com/opensalt/opensalt/blob/db6c0c100a8509155a26f5db1d43c9802384d2af/src/Salt/SiteBundle/Controller/CommentsController.php#L34)):*
>         $comment->setCreatedAt(new \DateTime($request->request->get('created')));
>         $comment->setUpdatedAt(new \DateTime($request->request->get('modified')));

These values should be locally generated, not just accepting whatever the user decides to send.

---

*[src/Salt/SiteBundle/Controller/CommentsController.php, line 88 at r1](https://reviewable.io:443/reviews/opensalt/opensalt/90#-Kpu-KknNkDhjnHK7RXk:-Kpu-KknNkDhjnHK7RXl:b-txf0ys) ([raw file](https://github.com/opensalt/opensalt/blob/db6c0c100a8509155a26f5db1d43c9802384d2af/src/Salt/SiteBundle/Controller/CommentsController.php#L88)):*
> ```PHP
>     /**
>      * @Route("/comments/delete/{id}")
>      * @Method("POST")
> ```

This shoudl probably use a `DELETE` method on `/comments/{id}` instead of a `POST` method.

---

*[src/Salt/SiteBundle/Controller/CommentsController.php, line 93 at r1](https://reviewable.io:443/reviews/opensalt/opensalt/90#-Kpu-_BDd3CZCaFkRpiW:-Kpu-_BDd3CZCaFkRpiX:babddh5) ([raw file](https://github.com/opensalt/opensalt/blob/db6c0c100a8509155a26f5db1d43c9802384d2af/src/Salt/SiteBundle/Controller/CommentsController.php#L93)):*
> ```PHP
>     {
>         if (!$comment) {
>             throw $this->createNotFoundException('Unable to find the comment');
> ```

If we need an error here, a `410 Gone` would make more sense than a `404 Not Found` for subsequent DELETEs.

---

*[src/Salt/SiteBundle/Entity/Comment.php, line 12 at r1](https://reviewable.io:443/reviews/opensalt/opensalt/90#-Kpu6DtcH-Assd3r6YvZ:-Kpu6DtcH-Assd3r6Yv_:b-i5t3e8) ([raw file](https://github.com/opensalt/opensalt/blob/db6c0c100a8509155a26f5db1d43c9802384d2af/src/Salt/SiteBundle/Entity/Comment.php#L12)):*
> ```PHP
>  *
>  * @ORM\Entity
>  * @ORM\Table(name="comments")
> ```

We might want to prefix the table name with some sort of namespace to keep the tables grouped by name.

The table name should be singular to keep the same convention the other tables are using.

---

*[src/Salt/SiteBundle/Entity/Comment.php, line 55 at r1](https://reviewable.io:443/reviews/opensalt/opensalt/90#-Kpu2Q80a0njYUBwLMJS:-Kpu2Q80a0njYUBwLMJT:bh5bhp5) ([raw file](https://github.com/opensalt/opensalt/blob/db6c0c100a8509155a26f5db1d43c9802384d2af/src/Salt/SiteBundle/Entity/Comment.php#L55)):*
> ```PHP
> 
>     /**
>      * @ORM\Column(type="boolean")
> ```

Does this make sense to be persisted to the database?

---

*[src/Salt/SiteBundle/Entity/Comment.php, line 65 at r1](https://reviewable.io:443/reviews/opensalt/opensalt/90#-Kpu46WSKMqbeomDTlSr:-Kpu46WSKMqbeomDTlSs:b-4ijlgh) ([raw file](https://github.com/opensalt/opensalt/blob/db6c0c100a8509155a26f5db1d43c9802384d2af/src/Salt/SiteBundle/Entity/Comment.php#L65)):*
> ```PHP
> 
>     /**
>      * @ORM\Column(type="boolean")
> ```

Does this make sense to persist to the database either?  Assuming it is similar to `createdByCurrentUser`.

---

*[src/Salt/SiteBundle/Entity/Comment.php, line 79 at r1](https://reviewable.io:443/reviews/opensalt/opensalt/90#-Kpu4NBcC1OnR62Bde40:-Kpu4NBcC1OnR62Bde41:b-g8059v) ([raw file](https://github.com/opensalt/opensalt/blob/db6c0c100a8509155a26f5db1d43c9802384d2af/src/Salt/SiteBundle/Entity/Comment.php#L79)):*
<details><summary><i alt="summary">Quoted 5 lines of code…</i></summary>
>     /**
>      * @ORM\Column(type="datetime")
>      * @Serializer\SerializedName("created")
>      */
>     private $createdAt;
</details>
>  
<details><summary><i alt="summary">Quoted 5 lines of code…</i></summary>
>     /**
>      * @ORM\Column(type="datetime")
>      * @Serializer\SerializedName("modified")
>      */
>     private $updatedAt;
</details>

Should probably make these automatically get the current time on the appropriate events instead of a value having to be passed in.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/opensalt/opensalt/90)*

JorgeLu1s:
Review status: all files reviewed at latest revision, 19 unresolved discussions, some commit checks failed.

---

*[app/Resources/views/base.html.twig, line 16 at r1](https://reviewable.io:443/reviews/opensalt/opensalt/90#-KptytLNWArOlGyniGQL:-KpvJLVcinaXGdrJjc4H:b-64o138) ([raw file](https://github.com/opensalt/opensalt/blob/db6c0c100a8509155a26f5db1d43c9802384d2af/app/Resources/views/base.html.twig#L16)):*

Should probably add `|asset_version` here as well to help with cache busting.
</blockquote></details>

so I just have to add `|asset_version` ?

---

*[src/CftfBundle/Resources/views/DocTree/view.html.twig, line 425 at r1](https://reviewable.io:443/reviews/opensalt/opensalt/90#-KptzAIrNp5fvRMZXbCT:-KpvJasZW-wveH5tQQZF:bo7k93v) ([raw file](https://github.com/opensalt/opensalt/blob/db6c0c100a8509155a26f5db1d43c9802384d2af/src/CftfBundle/Resources/views/DocTree/view.html.twig#L425)):*

This might need `|asset_version` to help with cache busting.
</blockquote></details>

just need to add `|asset_version` ?

---

*[src/Salt/SiteBundle/Controller/CommentsController.php, line 26 at r1](https://reviewable.io:443/reviews/opensalt/opensalt/90#-Kpu1mtJYuTLdDcfRzWr:-KpvJgYnDhiJQqM2_fdU:b-saq1ea) ([raw file](https://github.com/opensalt/opensalt/blob/db6c0c100a8509155a26f5db1d43c9802384d2af/src/Salt/SiteBundle/Controller/CommentsController.php#L26)):*

Why a `0` instead of `null`?  The 0 gets turned into null when listing the comment below, so does 0 mean anything?

Can `parent` be a FK?  Or does the field have additional structure?
</blockquote></details>

the parent is just in case of a reply comment, the `parent` column would be the `id` of the parent comment

---

*[src/Salt/SiteBundle/Controller/CommentsController.php, line 30 at r1](https://reviewable.io:443/reviews/opensalt/opensalt/90#-Kpu3j7UAXRTb8MAe4Xk:-KpvK4mTt042sKvtlFQt:b4jmham) ([raw file](https://github.com/opensalt/opensalt/blob/db6c0c100a8509155a26f5db1d43c9802384d2af/src/Salt/SiteBundle/Controller/CommentsController.php#L30)):*

Would `upvoteCount` ever not be 0 when the comment is initially created?  Is an arbitrary value okay to accept here?
</blockquote></details>

yes I think we don't need to set the upvoteCount here, actually we are not implementing this feature

---

*[src/Salt/SiteBundle/Controller/CommentsController.php, line 31 at r1](https://reviewable.io:443/reviews/opensalt/opensalt/90#-Kpu55epyq6vnI6VK6xN:-KpvKRGduA-dwCGWJ-nZ:b-78fvyr) ([raw file](https://github.com/opensalt/opensalt/blob/db6c0c100a8509155a26f5db1d43c9802384d2af/src/Salt/SiteBundle/Controller/CommentsController.php#L31)):*

This should probably be set as the default in the entity, so wouldn't have to set explicitly here.
</blockquote></details>

yes, we can ignore this because we are not implementing the upvote feature until now, maybe we can make the column nullable, what do you think?

---

*[src/Salt/SiteBundle/Controller/CommentsController.php, line 32 at r1](https://reviewable.io:443/reviews/opensalt/opensalt/90#-KptzP2XPBoOxwPN_0vd:-KpvMN7AcczoYMzMWJme:b-pejywz) ([raw file](https://github.com/opensalt/opensalt/blob/db6c0c100a8509155a26f5db1d43c9802384d2af/src/Salt/SiteBundle/Controller/CommentsController.php#L32)):*

We might want to separate the components of the `item` reference with something like a `:`, or possibly even use a real FK.

There is also no verification that such an itemType + itemId exists here, nor that the user has access to view it.
</blockquote></details>

Currently we are adding comments to docs and items, we would need a FK to both of them, right? what do you think would be better? add the `:` or the FK?

The `itemType` is the type of the node, can be 'document' or 'item' I think this parameter always will be present, but yes, we can add the validation here.

The `itemId` is the id of the node (document or item), so I think this parameter always will be present.

right now everybody can see the comments, even if you don't have an account 

---

*[src/Salt/SiteBundle/Controller/CommentsController.php, line 34 at r1](https://reviewable.io:443/reviews/opensalt/opensalt/90#-Kpu2sLubXEV698QoDVd:-KpvNoPCTkiaDv3bL41T:b-6lxrwj) ([raw file](https://github.com/opensalt/opensalt/blob/db6c0c100a8509155a26f5db1d43c9802384d2af/src/Salt/SiteBundle/Controller/CommentsController.php#L34)):*

>         $comment->setCreatedAt(new \DateTime($request->request->get('created')));
>         $comment->setUpdatedAt(new \DateTime($request->request->get('modified')));

These values should be locally generated, not just accepting whatever the user decides to send.
</blockquote></details>

you're right, should I just create a new `DateTime` object and format it?

---

*[src/Salt/SiteBundle/Controller/CommentsController.php, line 88 at r1](https://reviewable.io:443/reviews/opensalt/opensalt/90#-Kpu-KknNkDhjnHK7RXk:-KpvOq_BH4tGmEWvAEgK:b-jhy1bs) ([raw file](https://github.com/opensalt/opensalt/blob/db6c0c100a8509155a26f5db1d43c9802384d2af/src/Salt/SiteBundle/Controller/CommentsController.php#L88)):*

This shoudl probably use a `DELETE` method on `/comments/{id}` instead of a `POST` method.
</blockquote></details>

yes! wanted to ask this to you but I totally forgot xD, I can't use `DELETE`, `PUT` or `PATCH`, do you know why?

---

*[src/Salt/SiteBundle/Controller/CommentsController.php, line 93 at r1](https://reviewable.io:443/reviews/opensalt/opensalt/90#-Kpu-_BDd3CZCaFkRpiW:-KpvPGCNwKU7HHNo9xmE:b-jj848j) ([raw file](https://github.com/opensalt/opensalt/blob/db6c0c100a8509155a26f5db1d43c9802384d2af/src/Salt/SiteBundle/Controller/CommentsController.php#L93)):*

If we need an error here, a `410 Gone` would make more sense than a `404 Not Found` for subsequent DELETEs.
</blockquote></details>

yes, you're right

---

*[src/Salt/SiteBundle/Entity/Comment.php, line 12 at r1](https://reviewable.io:443/reviews/opensalt/opensalt/90#-Kpu6DtcH-Assd3r6YvZ:-KpvPMw8ucxA7vRfo_Uc:b-149bmz) ([raw file](https://github.com/opensalt/opensalt/blob/db6c0c100a8509155a26f5db1d43c9802384d2af/src/Salt/SiteBundle/Entity/Comment.php#L12)):*

We might want to prefix the table name with some sort of namespace to keep the tables grouped by name.

The table name should be singular to keep the same convention the other tables are using.
</blockquote></details>

go it, any suggestion?

---

*[src/Salt/SiteBundle/Entity/Comment.php, line 55 at r1](https://reviewable.io:443/reviews/opensalt/opensalt/90#-Kpu2Q80a0njYUBwLMJS:-KpvPS_6EwQG5SljqGJt:b-n7ooio) ([raw file](https://github.com/opensalt/opensalt/blob/db6c0c100a8509155a26f5db1d43c9802384d2af/src/Salt/SiteBundle/Entity/Comment.php#L55)):*

Does this make sense to be persisted to the database?
</blockquote></details>

no, I just added this because I need to be able of set it to false or true when retrieving the comments, how do I to not persist it in the database?

---

*[src/Salt/SiteBundle/Entity/Comment.php, line 65 at r1](https://reviewable.io:443/reviews/opensalt/opensalt/90#-Kpu46WSKMqbeomDTlSr:-KpvPhy5_DgzIcM5g9G_:b4uaoku) ([raw file](https://github.com/opensalt/opensalt/blob/db6c0c100a8509155a26f5db1d43c9802384d2af/src/Salt/SiteBundle/Entity/Comment.php#L65)):*

Does this make sense to persist to the database either?  Assuming it is similar to `createdByCurrentUser`.
</blockquote></details>

yes, not needed to be persisted in the database, we are not implementing the upvote feature for now, maybe in a short future, what do you think we could make with this?

---

*[src/Salt/SiteBundle/Entity/Comment.php, line 79 at r1](https://reviewable.io:443/reviews/opensalt/opensalt/90#-Kpu4NBcC1OnR62Bde40:-KpvPw91CL8c0k5-PTb4:bqkbs8v) ([raw file](https://github.com/opensalt/opensalt/blob/db6c0c100a8509155a26f5db1d43c9802384d2af/src/Salt/SiteBundle/Entity/Comment.php#L79)):*

>     /**
>      * @ORM\Column(type="datetime")
>      * @Serializer\SerializedName("created")
>      */
>     private $createdAt;
>  
>     /**
>      * @ORM\Column(type="datetime")
>      * @Serializer\SerializedName("modified")
>      */
>     private $updatedAt;

Should probably make these automatically get the current time on the appropriate events instead of a value having to be passed in.
</blockquote></details>

hmm I think this answers my previous question about the `DateTime` 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/opensalt/opensalt/90)*

roverwolf:
Reviewed 10 of 10 files at r3.
Review status: all files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---

*[app/DoctrineMigrations/Version20170803164313.php, line 56 at r3](https://reviewable.io:443/reviews/opensalt/opensalt/90#-Kqz8h8Iv7_rMICeJDC1:-Kqz8h8Iv7_rMICeJDC2:byxsvea) ([raw file](https://github.com/opensalt/opensalt/blob/fae05a9041694200d67e5046eb654a9f9e278ad7/app/DoctrineMigrations/Version20170803164313.php#L56)):*
> ```PHP
>         $this->addSql('ALTER TABLE ls_item ADD type VARCHAR(60) DEFAULT NULL COLLATE utf8_unicode_ci, CHANGE updated_at updated_at DATETIME DEFAULT CURRENT_TIMESTAMP NOT NULL');
>         $this->addSql('ALTER TABLE salt_user CHANGE locked locked TINYINT(1) DEFAULT \'0\' NOT NULL');
>     }
> ```

There are still a lot of `ALTER TABLE` SQL lines that have nothing to do with the tables you are creating here (in both `up` and `down`).

---

*[app/Resources/views/base.html.twig, line 16 at r1](https://reviewable.io:443/reviews/opensalt/opensalt/90#-KptytLNWArOlGyniGQL:-Kqz9Aqy6MJWytqBEDpg:bb77ufg) ([raw file](https://github.com/opensalt/opensalt/blob/db6c0c100a8509155a26f5db1d43c9802384d2af/app/Resources/views/base.html.twig#L16)):*
<details><summary><i>Previously, JorgeLu1s (Jorge Luis) wrote…</i></summary><blockquote>

so I just have to add `|asset_version` ?
</blockquote></details>

Yes.

---

*[src/Salt/SiteBundle/Controller/CommentsController.php, line 32 at r1](https://reviewable.io:443/reviews/opensalt/opensalt/90#-KptzP2XPBoOxwPN_0vd:-Kqz9uYm09JYAWS9rxTw:b-lnlfku) ([raw file](https://github.com/opensalt/opensalt/blob/db6c0c100a8509155a26f5db1d43c9802384d2af/src/Salt/SiteBundle/Controller/CommentsController.php#L32)):*
<details><summary><i>Previously, JorgeLu1s (Jorge Luis) wrote…</i></summary><blockquote>

Currently we are adding comments to docs and items, we would need a FK to both of them, right? what do you think would be better? add the `:` or the FK?

The `itemType` is the type of the node, can be 'document' or 'item' I think this parameter always will be present, but yes, we can add the validation here.

The `itemId` is the id of the node (document or item), so I think this parameter always will be present.

right now everybody can see the comments, even if you don't have an account 

</blockquote></details>

You are still not validating that the content of itemType or itemId are valid.  It might not matter, but will make it harder to remove junk if someone sends values (like `comment` or `43`) here.

---

*[src/Salt/SiteBundle/Controller/CommentsController.php, line 113 at r3](https://reviewable.io:443/reviews/opensalt/opensalt/90#-KqzC9RUcf0EQvAYpaJQ:-KqzC9RUcf0EQvAYpaJR:bmgj9nc) ([raw file](https://github.com/opensalt/opensalt/blob/fae05a9041694200d67e5046eb654a9f9e278ad7/src/Salt/SiteBundle/Controller/CommentsController.php#L113)):*
> ```PHP
> 
>         $comment->setUpvoteCount(
>             $comment->getUpvoteCount() + 1
> ```

You have a race condition here.

Would be better to either use an `UPDATE` DQL statement in the repository for `Comment` or better yet just fetch the count of related `CommentUpvote` records when fetching the comments (having the CommentUpvotes as a collection of associated objects to Comment).

---

*[src/Salt/SiteBundle/Entity/Comment.php, line 55 at r1](https://reviewable.io:443/reviews/opensalt/opensalt/90#-Kpu2Q80a0njYUBwLMJS:-KqzB3Q8JkW8nx0jwLby:b-b5car3) ([raw file](https://github.com/opensalt/opensalt/blob/db6c0c100a8509155a26f5db1d43c9802384d2af/src/Salt/SiteBundle/Entity/Comment.php#L55)):*
<details><summary><i>Previously, JorgeLu1s (Jorge Luis) wrote…</i></summary><blockquote>

no, I just added this because I need to be able of set it to false or true when retrieving the comments, how do I to not persist it in the database?
</blockquote></details>

It would not be persisted if you don't add the `@ORM` definition.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/opensalt/opensalt/90)*

roverwolf:
Reviewed 1 of 10 files at r3, 7 of 7 files at r4.
Review status: all files reviewed at latest revision, 9 unresolved discussions, some commit checks failed.

---

*[app/DoctrineMigrations/Version20170811154539.php, line 35 at r4](https://reviewable.io:443/reviews/opensalt/opensalt/90#-KrI-amFbDr9FRx-XzOi:-KrI-amG2UayL60GZAMy:b-7pme28) ([raw file](https://github.com/opensalt/opensalt/blob/53a567d0246cc24b9c72f0a32830c37adc955c01/app/DoctrineMigrations/Version20170811154539.php#L35)):*
> ```PHP
>         $this->abortIf($this->connection->getDatabasePlatform()->getName() !== 'mysql', 'Migration can only be executed safely on \'mysql\'.');
> 
>         $this->addSql('ALTER TABLE salt_comment_upvote DROP FOREIGN KEY FK_4DB1D19CF8697D13');
> ```

I'm wondering if the second FK also needs to be dropped.

---

*[app/Resources/assets/js/lsdoc/index.js, line 576 at r4](https://reviewable.io:443/reviews/opensalt/opensalt/90#-KrI-tPZHaxpk1absHIW:-KrI-tPZHaxpk1absHIX:b-jnxbhb) ([raw file](https://github.com/opensalt/opensalt/blob/53a567d0246cc24b9c72f0a32830c37adc955c01/app/Resources/assets/js/lsdoc/index.js#L576)):*
> ```JavaScript
>             putComment: function(commentJSON, success, error) {
>                 $.ajax({
>                     type: 'update',
> ```

I don't think `update` is a valid http verb.

---

*[src/Salt/SiteBundle/Controller/CommentsController.php, line 28 at r4](https://reviewable.io:443/reviews/opensalt/opensalt/90#-KrI0vpGVXXvxR03He7v:-KrI0vpGVXXvxR03He7w:bff1par) ([raw file](https://github.com/opensalt/opensalt/blob/53a567d0246cc24b9c72f0a32830c37adc955c01/src/Salt/SiteBundle/Controller/CommentsController.php#L28)):*
> ```PHP
>         $types = array('document', 'item');
> 
>         if (filter_var($itemId, FILTER_VALIDATE_INT) && in_array($itemType, $types)) {
> ```

Should this also do a `find($itemId)` on the $itemType's repository to ensure that the entity exists?

---

*[src/Salt/SiteBundle/Controller/CommentsController.php, line 31 at r4](https://reviewable.io:443/reviews/opensalt/opensalt/90#-KrI0guazcUB3E0QryUN:-KrI0guazcUB3E0QryUO:b-d66xlb) ([raw file](https://github.com/opensalt/opensalt/blob/53a567d0246cc24b9c72f0a32830c37adc955c01/src/Salt/SiteBundle/Controller/CommentsController.php#L31)):*
> ```PHP
> 
>             $comment->setContent(trim($request->request->get('content')));
>             $comment->setCommentId($request->request->get('id'));
> ```

Does this value actually mean anything?  What happens if a duplicate is added?

---

*[src/Salt/SiteBundle/Controller/CommentsController.php, line 33 at r4](https://reviewable.io:443/reviews/opensalt/opensalt/90#-KrI0dZShBp90qU058sI:-KrI0dZShBp90qU058sJ:bjv8h1t) ([raw file](https://github.com/opensalt/opensalt/blob/53a567d0246cc24b9c72f0a32830c37adc955c01/src/Salt/SiteBundle/Controller/CommentsController.php#L33)):*
> ```PHP
>             $comment->setCommentId($request->request->get('id'));
>             $comment->setParent(
>                 empty($request->request->get('parent'))?null:$request->request->get('parent')
> ```

This should probably also be validated.

---

*[src/Salt/SiteBundle/Controller/CommentsController.php, line 72 at r4](https://reviewable.io:443/reviews/opensalt/opensalt/90#-KrI1_5tT77Rg5-X7gih:-KrI1_5tT77Rg5-X7gii:b-fi5m90) ([raw file](https://github.com/opensalt/opensalt/blob/53a567d0246cc24b9c72f0a32830c37adc955c01/src/Salt/SiteBundle/Controller/CommentsController.php#L72)):*
> ```PHP
>                     $upvoteUser = $upvote->getUser();
>                     if ($upvoteUser->getId() == $user->getId()) {
>                         $comment->setUserHasUpvoted(($comment->getUpvoteCount() > 0)?true:false);
> ```

Do we need the `($comment->getUpvoteCount()...)` part here?  By definition won't `getUpvoteCount()` be greater than 0 (otherwise there would be no upvotes for the foreach to iterate over).

---

*[src/Salt/SiteBundle/Controller/CommentsController.php, line 84 at r4](https://reviewable.io:443/reviews/opensalt/opensalt/90#-KrI1p2T2g3T5bg-9s9O:-KrI1p2T2g3T5bg-9s9P:b-9r459a) ([raw file](https://github.com/opensalt/opensalt/blob/53a567d0246cc24b9c72f0a32830c37adc955c01/src/Salt/SiteBundle/Controller/CommentsController.php#L84)):*
> ```PHP
>     /**
>      * @Route("/comments/{id}")
>      * @Method("UPDATE")
> ```

UPDATE is not a valid HTTP verb.

Maybe you want PUT here?

GET, HEAD, POST, PUT, DELETE, CONNECT, OPTIONS, TRACE, PATCH
Ref: https://tools.ietf.org/html/rfc7231#section-4

---

*[src/Salt/SiteBundle/Entity/Comment.php, line 43 at r4](https://reviewable.io:443/reviews/opensalt/opensalt/90#-KrI2t3jASvGaRGnhbg_:-KrI2t3jASvGaRGnhbga:b-ip7jlg) ([raw file](https://github.com/opensalt/opensalt/blob/53a567d0246cc24b9c72f0a32830c37adc955c01/src/Salt/SiteBundle/Entity/Comment.php#L43)):*
> ```PHP
>      * @Serializer\Exclude()
>      */
>     private $userId;
> ```

You might want to change this to be a reference to the `User` object the same way as `CommentUpvote` is now using.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/opensalt/opensalt/90)*

roverwolf:
Reviewed 6 of 6 files at r5.
Review status: all files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[src/Salt/SiteBundle/Controller/CommentsController.php, line 73 at r5](https://reviewable.io:443/reviews/opensalt/opensalt/90#-KrblxEhpTo-vJO0l0RS:-KrblxEhpTo-vJO0l0RT:b-bgornd) ([raw file](https://github.com/opensalt/opensalt/blob/59784f668c69ab74b973e4773df445e5ffb9f72f/src/Salt/SiteBundle/Controller/CommentsController.php#L73)):*
> ```PHP
>                 foreach ($upvotes as $upvote) {
>                     if ($upvote->getUser()->getId() == $user->getId()) {
>                         $comment->setUserHasUpvoted(true);
> ```

Might want to add a `break;` here as once it has been determined that the user has upvoted then there is no need to continue checking the rest of the upvotes.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/opensalt/opensalt/90)*

roverwolf:
Reviewed 9 of 9 files at r6.
Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[app/DoctrineMigrations/Version20170811231606.php, line 35 at r4](https://reviewable.io:443/reviews/opensalt/opensalt/90#-KrI-amFbDr9FRx-XzOi:-Krg1wWSjZF23vc3tMPN:b-xd3t98) ([raw file](https://github.com/opensalt/opensalt/blob/53a567d0246cc24b9c72f0a32830c37adc955c01/app/DoctrineMigrations/Version20170811231606.php#L35)):*

I'm wondering if the second FK also needs to be dropped.
</blockquote></details>

As well as the FK in salt_comment.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/opensalt/opensalt/90)*

roverwolf:
Review status: all files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[src/Salt/SiteBundle/Controller/CommentsController.php, line 26 at r6](https://reviewable.io:443/reviews/opensalt/opensalt/90#-Krg39zhfoO84ndtOz40:-Krg39zhfoO84ndtOz41:b-iuyhsc) ([raw file](https://github.com/opensalt/opensalt/blob/7be72d2981df6222f733d92aedf89dc155f3b231/src/Salt/SiteBundle/Controller/CommentsController.php#L26)):*
> ```PHP
>     {
>         $comment = new Comment();
>         $user = $this->getUser();
> ```

Are anonymous users allowed to create comments?

If so, there will be issues below when referencing `$user` to get the username and org.
If not, then the action (and other actions) should probably be restricted to logged in users.

---

*[src/Salt/SiteBundle/Entity/Comment.php, line 159 at r6](https://reviewable.io:443/reviews/opensalt/opensalt/90#-Krg2mQYFYQ5hSSDL7qF:-Krg2mQZVOGMRHWYkaEJ:b-17xnuz) ([raw file](https://github.com/opensalt/opensalt/blob/7be72d2981df6222f733d92aedf89dc155f3b231/src/Salt/SiteBundle/Entity/Comment.php#L159)):*
> ```PHP
>      * Set user
>      *
>      * @param \Salt\UserBundle\Entity\User $userId
> ```

The param should now reference `$user` not `$userId`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/opensalt/opensalt/90)*

brandonopened:
Re: Anonymous - the idea originally was anonymous users needing to input their name and org. Jorge and I had talked about implementing a new user type for env variable but I am not 100% sure if that was done. (If so we need an easy way to create usernames). If not then yes this should be addressed and the Comments needs new fields added - org and email as we need to be able to contact folks who make comments later.

roverwolf:
Review status: 13 of 15 files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---

*[src/Salt/SiteBundle/Controller/CommentsController.php, line 26 at r7](https://reviewable.io:443/reviews/opensalt/opensalt/90#-KrkOAqOd8TdU47ORpMr:-KrkOAqOd8TdU47ORpMs:b-b2xca7) ([raw file](https://github.com/opensalt/opensalt/blob/173cc23242d82efe2041b350a16af16da3f36088/src/Salt/SiteBundle/Controller/CommentsController.php#L26)):*
> $user = $this->getUser();

Getting the user should probably be changed to being an argument to the method using the user resolver.

`public function newAction(Request $request, UserInterface $user)`

[Ref for new User value resolver](http://symfony.com/blog/new-in-symfony-3-2-user-value-resolver-for-controllers)

---

*[src/Salt/SiteBundle/Controller/CommentsController.php, line 29 at r7](https://reviewable.io:443/reviews/opensalt/opensalt/90#-KrkMcbkIjDkZ630m2Gv:-KrkMcbkIjDkZ630m2Gw:b1bhoyk) ([raw file](https://github.com/opensalt/opensalt/blob/173cc23242d82efe2041b350a16af16da3f36088/src/Salt/SiteBundle/Controller/CommentsController.php#L29)):*
> if ($user) {

Better would be to use an `@Security("has_role('ROLE_USER')")`annotation on the method instead of checking the `$user` exists.

Better yet would be to add a voter (similar to the `FrameworkAccessVoter`) to allow for a specific check on allowing comment view/add (using something like `@Security("is_granted('comment_view')")`).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/opensalt/opensalt/90)*

roverwolf:
Reviewed 1 of 2 files at r7, 3 of 3 files at r8.
Review status: all files reviewed at latest revision, 2 unresolved discussions, some commit checks broke.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/opensalt/opensalt/90)*

roverwolf:
Reviewed 4 of 4 files at r9.
Review status: all files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/opensalt/opensalt/90)*

app/Resources/assets/js/lsdoc/index.js
```diff
@@ -556,6 +556,134 @@ function listRepositories(){
     $('#back').html('');
 }
 
+var CommentSystem = (function(){
```
roverwolf: Should this be in a separate JS file loaded only if commenting is turned on?

app/Resources/views/base.html.twig
```diff
@@ -14,6 +14,7 @@
         <title>{% block title %}OpenSALT{% endblock %}</title>
         {% block stylesheets %}
             <link rel="stylesheet" href="{{ asset('css/main.css'|asset_version) }}" />
+            <link rel="stylesheet" href="{{ asset('css/jquery-comments.css'|asset_version) }}" />
```
roverwolf: This should probably be wrapped to only be loaded if commenting is enabled.

src/Salt/SiteBundle/Controller/CommentsController.php
```diff
@@ -0,0 +1,215 @@
...
+        $parentId = $request->request->get('parent');
+
+        if ($this->existItem($itemId, $itemType)) {
+            $comment->setContent(trim($request->request->get('content')));
+            $comment->setUser($user);
+            $comment->setFullname($user->getUsername().' - '.$user->getOrg()->getName());
+            $comment->setItem($itemType.':'.$itemId);
+
+            if (!empty($parentId) && filter_var($parentId, FILTER_VALIDATE_INT)) {
+                $parent = $em->getRepository('SaltSiteBundle:Comment')->findById($parentId);
```
roverwolf: This should probably be `->find($parentId)` instead of `->findById($parentId)`.

src/Salt/SiteBundle/Controller/CommentsController.php
```diff
@@ -0,0 +1,215 @@
...
+                $comment->setParent(($parent)?$parentId:null);
```
roverwolf: Should `parent` be a Comment object instead of an integer?

src/Salt/SiteBundle/Controller/CommentsController.php
```diff
@@ -0,0 +1,215 @@
...
+     *
+     * @Method("PUT")
+     *
+     * @Security("is_granted('comment')")
+     */
+    public function updateAction(Comment $comment, Request $request, UserInterface $user)
+    {
+        $em = $this->getDoctrine()->getManager();
+
+        $comment->setContent($request->request->get('content'));
```
roverwolf: Can any user update the comment, or only the author?

src/Salt/SiteBundle/Controller/CommentsController.php
```diff
@@ -0,0 +1,215 @@
...
+     *
+     * @Method("DELETE")
+     *
+     * @Security("is_granted('comment')")
+     */
+    public function deleteAction(Comment $comment, UserInterface $user)
+    {
+        $em = $this->getDoctrine()->getManager();
+
+        $em->remove($comment);
```
roverwolf: Can any user remove the comment, or does the `$comment->getUser()` need to match `$user`?

src/Salt/SiteBundle/Controller/CommentsController.php
```diff
@@ -0,0 +1,215 @@
...
+    }
+
+    private function existItem($itemId, $itemType)
+    {
+        if (filter_var($itemId, FILTER_VALIDATE_INT)) {
+            $em = $this->getDoctrine()->getManager();
+
+            switch ($itemType) {
+                case 'document':
+                    $item = $em->getRepository('CftfBundle:LsDoc')->findOneById($itemId);
```
roverwolf: This should probably be just `->find($itemId)`

src/Salt/SiteBundle/Controller/CommentsController.php
```diff
@@ -0,0 +1,215 @@
...
+                    break;
+                case 'item':
+                    $item = $em->getRepository('CftfBundle:LsItem')->findOneById($itemId);
```
roverwolf: This should probably be just `->find($itemId)` as well.

src/Salt/SiteBundle/Controller/CommentsController.php
```diff
@@ -0,0 +1,215 @@
...
+    {
+        return $this->get('jms_serializer')
+            ->serialize($data, 'json');
+    }
+
+    private function apiResponse($data, $statusCode = 200)
+    {
+        $json = $this->serialize($data);
+
+        return new Response($json, $statusCode, [
```
roverwolf: This should probably create a `JsonResponse` instead of a `Response` (and then no need to set the additional header).

src/Salt/SiteBundle/Entity/Comment.php
```diff
@@ -0,0 +1,367 @@
...
+     * @ORM\Id
+     * @ORM\GeneratedValue(strategy="AUTO")
+     * @ORM\Column(type="integer")
+     */
+    private $id;
+
+    /**
+     * @ORM\Column(type="integer", nullable=true)
+     */
+    private $parent;
```
roverwolf: Might want to change this to be a `ManyToOne` referencing a `Comment` instead of just an int.

src/Salt/SiteBundle/Entity/Comment.php
```diff
@@ -0,0 +1,367 @@
...
+    }
+
+    /**
+     * Set id
+     *
+     * @param int $id
+     *
+     * @return Comment
+     */
+    public function setId($id)
```
roverwolf: `$id` is auto-generated, so probably makes sense to remove this method.

src/Salt/SiteBundle/Entity/Comment.php
```diff
@@ -0,0 +1,367 @@
...
+     */
+    public function getFullname()
+    {
+        return $this->fullname;
+    }
+
+    /**
+     * Set createdByCurrentUser
+     *
+     * @param string $createdByCurrentUser
```
roverwolf: Is this a string or a bool?

src/Salt/SiteBundle/Controller/CommentsController.php
```diff
@@ -0,0 +1,215 @@
...
+            }
+
+            $em->persist($comment);
+            $em->flush();
+
+            $response = $this->apiResponse($comment);
+            return $response;
+        }
+
+        return new Response('Item not found', 404);
```
roverwolf: Should this be a JsonResponse like the apiResponse creates?

app/Resources/assets/js/lsdoc/index.js
JorgeLu1s: How can I move this section in a different JS file and load it when the commenting is turned on? do I have to make some change in the gulpfile?

app/Resources/views/base.html.twig
JorgeLu1s: done

src/Salt/SiteBundle/Controller/CommentsController.php
JorgeLu1s: done

src/Salt/SiteBundle/Controller/CommentsController.php
JorgeLu1s: done

src/Salt/SiteBundle/Controller/CommentsController.php
JorgeLu1s: done

src/Salt/SiteBundle/Controller/CommentsController.php
JorgeLu1s: done

src/Salt/SiteBundle/Controller/CommentsController.php
JorgeLu1s: done

src/Salt/SiteBundle/Controller/CommentsController.php
JorgeLu1s: done

src/Salt/SiteBundle/Controller/CommentsController.php
JorgeLu1s: done

src/Salt/SiteBundle/Entity/Comment.php
JorgeLu1s: done

src/Salt/SiteBundle/Entity/Comment.php
JorgeLu1s: done

src/Salt/SiteBundle/Entity/Comment.php
JorgeLu1s: it's a bool, done

src/Salt/SiteBundle/Controller/CommentsController.php
JorgeLu1s: I'm going to make the change, although the only thing I need from this response is the status code.

roverwolf:
Reviewed 1 of 7 files at r5, 9 of 9 files at r10.
Review status: 15 of 17 files reviewed at latest revision, 1 unresolved discussion.

---

*[app/Resources/views/base.html.twig, line 18 at r10](https://reviewable.io:443/reviews/opensalt/opensalt/90#-KsG6d9PG0RIP5QtZK6F:-KsG6d9PG0RIP5QtZK6G:b-ftmsl6) ([raw file](https://github.com/opensalt/opensalt/blob/cc9acdaa459a7718169efc61954828be362ffaca/app/Resources/views/base.html.twig#L18)):*
> ```Twig
>             <link rel="stylesheet" href="{{ asset('css/main.css'|asset_version) }}" />
>             {% if 'comments' is active feature %}
>                 <link rel="stylesheet" href="{{ asset('css/comments.css'|asset_version) }}" />
> ```

If this is only used in `view.html.twig` (as that is where comment.js is loaded) then maybe move the loading of the css there as well?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/opensalt/opensalt/90)*

JorgeLu1s:
Review status: 15 of 17 files reviewed at latest revision, 1 unresolved discussion.

---

*[app/Resources/views/base.html.twig, line 18 at r10](https://reviewable.io:443/reviews/opensalt/opensalt/90#-KsG6d9PG0RIP5QtZK6F:-KsGAimTRyuYn7Fezt-4:bjkaiu7) ([raw file](https://github.com/opensalt/opensalt/blob/cc9acdaa459a7718169efc61954828be362ffaca/app/Resources/views/base.html.twig#L18)):*

If this is only used in `view.html.twig` (as that is where comment.js is loaded) then maybe move the loading of the css there as well?
</blockquote></details>

Can I add the `{% block stylesheets %}` anywhere in `view.html.twig`? or should I overwrite/extends the `<head>` from `base.html.twig`?  

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/opensalt/opensalt/90)*

roverwolf:
Review status: 13 of 16 files reviewed at latest revision, 1 unresolved discussion.

---

*[app/Resources/views/base.html.twig, line 18 at r10](https://reviewable.io:443/reviews/opensalt/opensalt/90#-KsG6d9PG0RIP5QtZK6F:-KsGUNUti5TBX7e6eevU:b-wp3uqo) ([raw file](https://github.com/opensalt/opensalt/blob/cc9acdaa459a7718169efc61954828be362ffaca/app/Resources/views/base.html.twig#L18)):*
<details><summary><i>Previously, JorgeLu1s (Jorge Luis) wrote…</i></summary><blockquote>

Can I add the `{% block stylesheets %}` anywhere in `view.html.twig`? or should I overwrite/extends the `<head>` from `base.html.twig`?  
</blockquote></details>

Yes, the `{% block stylesheets %}` can be put anywhere in `view.html.twig` but needs `{{ parent() }}` in it so that the lines from the original is there as well.

I've made the change and committed it to this branch.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/opensalt/opensalt/90)*

roverwolf:
Reviewed 2 of 2 files at r11, 1 of 2 files at r12.
Review status: all files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/opensalt/opensalt/90)*

roverwolf:
Reviewed 4 of 5 files at r13, 2 of 2 files at r14.
Review status: all files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/opensalt/opensalt/90)*

roverwolf:
Reviewed 8 of 8 files at r15.
Review status: all files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/opensalt/opensalt/90)*

roverwolf:
This is for issue #112

roverwolf:
Review status: all files reviewed at latest revision, 8 unresolved discussions, all commit checks successful.

---

*[tests/_support/AcceptanceTester.php, line 30 at r15](https://reviewable.io:443/reviews/opensalt/opensalt/90#-KtSnPDhv-orbyeZGh4O:-KtSnPDiS9DT_UNwwFFS:b-kh5fkj) ([raw file](https://github.com/opensalt/opensalt/blob/8e859153e907badc7c3aec624b2f587a10dfaa80/tests/_support/AcceptanceTester.php#L30)):*
> ```PHP
>     private $users = [];
>     private $lastUser = null;
>     private $lsDoc = null;
> ```

`$lsDoc` isn't used, but `$lsDocId` is referenced and not defined.

---

*[tests/_support/AcceptanceTester.php, line 182 at r15](https://reviewable.io:443/reviews/opensalt/opensalt/90#-KtSnkd4ZvKmmkxxvvCZ:-KtSnkd4ZvKmmkxxvvC_:buzx4ty) ([raw file](https://github.com/opensalt/opensalt/blob/8e859153e907badc7c3aec624b2f587a10dfaa80/tests/_support/AcceptanceTester.php#L182)):*
> ```PHP
>      public function getDocId()
>      {
>          return $this->lsDocId;
> ```

This might not be defined, maybe it should check and call `getLastFrameworkId()` if not defined?

---

*[tests/acceptance/CommentCest.php, line 4 at r15](https://reviewable.io:443/reviews/opensalt/opensalt/90#-KtT-yJO0iP-tfKc9xQJ:-KtT-yJPa8gFCc0g7gkm:b-aap1hv) ([raw file](https://github.com/opensalt/opensalt/blob/8e859153e907badc7c3aec624b2f587a10dfaa80/tests/acceptance/CommentCest.php#L4)):*
> ```PHP
> 
> 
> class CommentCest
> ```

Should this do any testing with item comments as well as a document?
Should we verify that a different user (anonymous or not) can see the changes?

---

*[tests/acceptance/CommentCest.php, line 8 at r15](https://reviewable.io:443/reviews/opensalt/opensalt/90#-KtSntl7rKq5eEe9lla-:-KtSntl7rKq5eEe9lla0:bolsanh) ([raw file](https://github.com/opensalt/opensalt/blob/8e859153e907badc7c3aec624b2f587a10dfaa80/tests/acceptance/CommentCest.php#L8)):*
> ```PHP
>     public function _before(AcceptanceTester $I)
> {
> }
> ```

Missing indent, but are these two methods even needed?

---

*[tests/acceptance/CommentCest.php, line 16 at r15](https://reviewable.io:443/reviews/opensalt/opensalt/90#-KtSo4ff40lRbtWxSKCy:-KtSo4ff40lRbtWxSKCz:b-i901sy) ([raw file](https://github.com/opensalt/opensalt/blob/8e859153e907badc7c3aec624b2f587a10dfaa80/tests/acceptance/CommentCest.php#L16)):*
> ```PHP
>     // tests
>     public function seeCommentsSectionAsAnAnonymousUser(AcceptanceTester $I)
>     {
> ```

Might want to ensure the user is not logged in first.
We probably need a way to skip these tests if the commenting feature is not enabled (posibly adding others to ensure that commenting isn't showing up).

---

*[tests/acceptance/CommentCest.php, line 36 at r15](https://reviewable.io:443/reviews/opensalt/opensalt/90#-KtSoXlIq5hX_AM_RP1A:-KtSoXlIq5hX_AM_RP1B:b-qxoz75) ([raw file](https://github.com/opensalt/opensalt/blob/8e859153e907badc7c3aec624b2f587a10dfaa80/tests/acceptance/CommentCest.php#L36)):*
> ```PHP
>         $I->amOnPage('/cftree/doc/'.$I->getDocId());
>         $I->loginAs('ROLE_EDITOR', 'Username', 'Password');
>         $I->click('.textarea');
> ```

Might want to be more specfic with the selector, `.textarea` is pretty generic and may apply to other things on the page at some point.

---

*[tests/acceptance/CommentCest.php, line 57 at r15](https://reviewable.io:443/reviews/opensalt/opensalt/90#-KtSpHE-ldeYmbm5fdUg:-KtSpHE-ldeYmbm5fdUh:b93zvyk) ([raw file](https://github.com/opensalt/opensalt/blob/8e859153e907badc7c3aec624b2f587a10dfaa80/tests/acceptance/CommentCest.php#L57)):*
> ```PHP
>         $I->amOnPage('/cftree/doc/'.$I->getDocId());
>         $I->loginAs('ROLE_EDITOR', 'Username', 'Password');
>         $upvotes = $I->grabTextFrom('.upvote');
> ```

Does this work when there are multiple comments on the page?  (same for other places where `.upvote` is used).

---

*[tests/acceptance/CommentCest.php, line 59 at r15](https://reviewable.io:443/reviews/opensalt/opensalt/90#-KtT-oGYItoDEaqe8Z4S:-KtT-oGYItoDEaqe8Z4T:b4xzn38) ([raw file](https://github.com/opensalt/opensalt/blob/8e859153e907badc7c3aec624b2f587a10dfaa80/tests/acceptance/CommentCest.php#L59)):*
> ```PHP
>         $upvotes = $I->grabTextFrom('.upvote');
>         $I->click('.upvote');
>         $I->wait(2);
> ```

Might want to use `$I->waitForElementChange()` to avoid the full 2 second delay.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/opensalt/opensalt/90)*

roverwolf:
Reviewed 2 of 5 files at r16, 6 of 6 files at r17, 3 of 7 files at r18, 7 of 8 files at r19.
Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/opensalt/opensalt/90)*

roverwolf:
Reviewed 5 of 5 files at r20.
Review status: all files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/opensalt/opensalt/90)*

roverwolf:
Reviewed 1 of 1 files at r21.
Review status: all files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/opensalt/opensalt/90)*

PR #94, opened 2017-08-11
share code between package export in UI and the API

### LINKED ISSUE #93: The "Export CF Package JSON" link in the UI should generate the same output as the API
Currently the link in the UI to export a CF Package does not give the same output as what the API provides.  They should be the same.

roverwolf:
Use same serialization code when exporting from the UI as the API uses.

resolves #93

---
This change is [Reviewable](https://reviewable.io/reviews/opensalt/opensalt/94)

### DIFF
--- a/src/CftfBundle/Controller/Api1Controller.php
+++ b/src/CftfBundle/Controller/Api1Controller.php
@@ -92,23 +92,7 @@ class Api1Controller extends Controller
             return $response;
         }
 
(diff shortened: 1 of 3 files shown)


roverwolf:
Marked as a break with backwards compatibility since the exported JSON will not match what it did before, but it now (correctly) matches what the API provides.

The export used by the javascript has not been changed.

Issue #112, opened 2017-08-30
Commenting on Frameworks


brandonopened:
- [x] #90 Write acceptance and unit tests for the framework feedback (Commenting system)
- [x] #132 Update documentation with how to enable the commenting feature

roverwolf:
User guide item moved to another issue.

PR #132, opened 2017-09-15
commenting_configuration

brandonopened:
OpenSALT uses the http://viima.github.io/jquery-comments/ bundle to allow editors to comment on and upvote Published 
frameworks.

 - env(COMMENTS_FEATURE): *default* inactive . *Enable* always-active

Correct formatting?

---
This change is [Reviewable](https://reviewable.io/reviews/opensalt/opensalt/132)


roverwolf:
This is to add configuration information for #112

brandonopened:
I see now I could/should have made these changes within that same PR and committed to that; will do that in the future.

PR #196, opened 2017-10-05
Feature/security

JosephLeon:
I added some code in nelmio security config for x-xss-protection. I also added code to enforce https.

---
This change is [Reviewable](https://reviewable.io/reviews/opensalt/opensalt/196)


app/config/security.yml
```diff
@@ -62,6 +62,6 @@ security:
     # http://symfony.com/doc/current/cookbook/security/access_control.htm
     access_control:
         #- { path: ^/login, roles: IS_AUTHENTICATED_ANONYMOUSLY, requires_channel: https }
-        - { path: ^/login, roles: IS_AUTHENTICATED_ANONYMOUSLY }
-        - { path: ^/logout, roles: IS_AUTHENTICATED_ANONYMOUSLY }
-        - { path: ^/, roles: IS_AUTHENTICATED_ANONYMOUSLY }
+        - { path: ^/login, roles: IS_AUTHENTICATED_ANONYMOUSLY, requires_channel: https }
+        - { path: ^/logout, roles: IS_AUTHENTICATED_ANONYMOUSLY, requires_channel: https }
+        - { path: ^/, roles: IS_AUTHENTICATED_ANONYMOUSLY, requires_channel: https }
```
roverwolf: There needs to be an easy way to turn this off via an environment variable for local development and automated tests.

roverwolf: Also note that this would be a backward compatibility break for developers, so should be documented to warn them.

app/config/config.yml
```diff
@@ -231,6 +232,20 @@ nelmio_security:
...
+    # Force ssl.
+    # forced_ssl: ~
+        # uncomment to whitelist certain URLs from having ssl forced.
+        # enabled: true
+        # whitelist:
+        #     - ^/unsecure/
+    # Recommended security setting for cookies when ssl is turned on.
+    # session:
+        # cookie_secure: true
+    # Forces Microsoft/Chrome XSS-Protection with its block mode.
```
roverwolf: In reverting the forced-SSL changes, the `xss_protection` block should be the only difference left.  Currenlty the move of where the `NelmioSecurityBundle` is loaded, the `requires_channel` changes in `security.yml`, and the duplicate `forced_ssl` block are still in the PR.

PR #211, opened 2017-10-13
Fix/ordering

JorgeLu1s:
This change is [Reviewable](https://reviewable.io/reviews/opensalt/opensalt/211)


app/Resources/assets/js/cftree/view-trees.js
```diff
@@ -290,8 +290,21 @@ apx.treeDocLoadCallback1 = function() {
                 return false;
             },
 
-            expand: function() {
+            expand: function(event, data) {
                 apx.treeDoc1.recordExpandedFolders(1);
+                data.node.sortChildren(function(a, b) {
```
roverwolf: This will override any explicit `sequenceNumber` values from the childOf LsAssociations.  We probably want those values to take precedence over the human coding scheme values.

This would also not sort the human coding scheme values [1.1, 1.2, 2.1, 2.2] correctly as it is only looking at the last segment.

PR #231, opened 2017-11-02
Truncated display name after the '@' in the comments section

JorgeLu1s:
This change is [Reviewable](https://reviewable.io/reviews/opensalt/opensalt/231)

### DIFF
--- a/src/Salt/SiteBundle/Entity/Comment.php
+++ b/src/Salt/SiteBundle/Entity/Comment.php
@@ -13,6 +13,12 @@ use Salt\UserBundle\Entity\User;
  *
  * @ORM\Entity(repositoryClass="CommentRepository")
  * @ORM\Table(name="salt_comment")
+ *
+ * @Serializer\VirtualProperty(
+ *     "fullname",
+ *     exp="object.getFullname()",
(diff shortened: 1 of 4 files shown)


src/Salt/SiteBundle/Entity/Comment.php
```diff
@@ -64,7 +64,10 @@ class Comment
     /**
      * @var string
      *
-     * @ORM\Column(type="string")
+     * @Serializer\Accessor(getter="getFullname")
+     * @Serializer\ReadOnly
+     * @Serializer\Type("string")
+     * @Serializer\SerializedName("fullname")
      */
     private $fullname;
```
roverwolf: There is no need for this field anymore, the `Serializer` annotations can be moved to be a [virtual property](http://jmsyst.com/libs/serializer/master/reference/annotations#virtualproperty).

roverwolf:
Note that instead of adding the annotation `@Serializer\VirtualProperty(...)` at the class level it could also have been added as just `@Serializer\VirtualProperty()` as the annotation on `getFullname()`.

Either way is fine, but less code is used if it was added to the function (but not necessarily as visible).

PR #265, opened 2017-11-23
Saved sequence number in the database after importing CSV file

JorgeLu1s:
I didn't add the tests because I don't know what to expect after the file is uploaded, I mean I'm not sure of what is the correct order in the tree.

---
This change is [Reviewable](https://reviewable.io/reviews/opensalt/opensalt/265)

### DIFF
--- a/src/CftfBundle/Entity/LsAssociation.php
+++ b/src/CftfBundle/Entity/LsAssociation.php
@@ -78,7 +78,6 @@ class LsAssociation extends AbstractLsBase implements CaseApiInterface
     public const PRECEDES = 'Precedes';
     public const SKILL_LEVEL = 'Has Skill Level';
     public const IS_PEER_OF = 'Is Peer Of';
-    public const SEQUENCE_NUMBER = 'Sequence Number';
 
     public const EXEMPLAR = 'Exemplar';
 
@@ -292,7 +291,6 @@ class LsAssociation extends AbstractLsBase implements CaseApiInterface
             'isRelatedTo' =>                  static::RELATED_TO,
             'replacedBy' =>                   static::REPLACED_BY,
             'hasSkillLevel' =>                static::SKILL_LEVEL,
-            'sequenceNumber' =>               static::SEQUENCE_NUMBER,
         ];
     }
 
--- a/src/CftfBundle/Entity/LsAssociation.php
+++ b/src/CftfBundle/Entity/LsAssociation.php
@@ -78,6 +78,7 @@ class LsAssociation extends AbstractLsBase implements CaseApiInterface
     public const PRECEDES = 'Precedes';
     public const SKILL_LEVEL = 'Has Skill Level';
     public const IS_PEER_OF = 'Is Peer Of';
+    public const SEQUENCE_NUMBER = 'Sequence Number';
 
     public const EXEMPLAR = 'Exemplar';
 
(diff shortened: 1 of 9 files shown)


src/CftfBundle/Entity/LsAssociation.php
```diff
@@ -291,6 +292,7 @@ public static function allTypesForImportFromCSV(){
             'isRelatedTo' =>                  static::RELATED_TO,
             'replacedBy' =>                   static::REPLACED_BY,
             'hasSkillLevel' =>                static::SKILL_LEVEL,
+            'sequenceNumber' =>               static::SEQUENCE_NUMBER,
```
roverwolf: `sequenceNumber` is not an association type

src/GithubFilesBundle/Service/GithubImport.php
```diff
@@ -165,10 +174,11 @@ public function saveAssociations($position, $content, $lsItemKeys, LsItem $lsIte
         $fieldsAndTypes = LsAssociation::allTypesForImportFromCSV();
         // We don't use is_child_of because that it alaready used to create parents relations before. :)
         // checking each association field
+        $sequenceNumber = $content[$position][trim($lsItemKeys['sequenceNumber'])];
         foreach ($fieldsAndTypes as $fieldName => $assocType){
```
roverwolf: Due to `sequenceNumber` being added to the `$fieldsAndTypes` array, this will try to create an association of type `sequenceNumber` if it finds data.  This is probably not what is wanted.

src/GithubFilesBundle/Service/GithubImport.php
```diff
@@ -165,10 +174,11 @@ public function saveAssociations($position, $content, $lsItemKeys, LsItem $lsIte
...
             if (array_key_exists($fieldName, $lsItemKeys) && $cfAssociations = $content[$position][trim($lsItemKeys[$fieldName])]) {
                 foreach (explode(',', $cfAssociations) as $cfAssociation) {
-                    $this->addItemRelated($lsDoc, $lsItem, $cfAssociation, $frameworkToAssociate, $assocType);
+                    $this->addItemRelated($lsDoc, $lsItem, $cfAssociation, $frameworkToAssociate, $assocType, $sequenceNumber);
```
roverwolf: Should all the associations be getting the `sequenceNumber` applied to them?  Or only the `isChildOf` ones?

Not that this looks like it will apply the **same** sequence number to all associations (exploded from `$cfAssociations`) which does not seem correct.

JorgeLu1s: well I don't know, I thought the sequence number should be applied to other items different of 'is Child of'. So are you saying that the sequence number should be imported just for the 'is Child Of' association type?

roverwolf: For this, I think so.

Otherwise if you had a set of `isRelatedTo` (or other) associations separated by commas they would all get the **same** sequence number.  I would assume that if they needed sequence numbers the values would need to be different for each of the associations that are the same type of association (for the same origin) for them to be useful.

brandonopened:
I can’t get screenshots right now but if on your local build look at the file... you should be able to see the top level statements based on the csv a bit? (Use human coding scheme as a reference...). If they aren’t in alphabetical order we are on the right track.

brandonopened:
Also i tend to agree this will probably only be used for Is child of.

Issue #270, opened 2017-11-30
org admins cannot import children


brandonopened:
Steps:
1 Login with Org admin
2 Click on Organization
3 Click on Framework from list
4 Verify import children is not appearing on framework display page

Actual:

Org Admin user is not able to see Import children button on framework display page

Expected:

Org Admin user should be able to see import children button and import through csv file.

  Note: Only superuser is able to see import children button, other role user like admin, superEditor, Editor and user are not able to see import children button (for more clarification please refer to attached video file.

sajeed42:
Hi Brandon,
Please find the attached video file

![ezgif com-video-to-gif](https://user-images.githubusercontent.com/33145683/33425622-3d14b0ea-d5e5-11e7-8aad-6d1a074ad066.gif)

roverwolf:
@brandonopened I think you were on the right track with #269.

 `createRights` is defined in `DocTreeController.php`,
the problem is that it is defined with `$authChecker->isGranted('create', $lsDoc)` which is not correct.

The create privilege takes a string argument like `lsdoc` as it is meant to determine if a user can create a specific type of object.  That check should be `$authChecker->isGranted('create', 'lsdoc')`.

But, for importing children or updating the framework, I think that having the edit privilege on the framework should be enough (I'm not sure why we would be checking for create rights there).

brandonopened:
Yeah, I felt it was perhaps too simple of a fix. I doubted myself about if we wanted an editor to be able to do that... but I feel like they may need too/doesn't clutter things too much for them to be honest. An org admin for sure should be able too, not sure how that slipped by previously.

roverwolf:
This has been fixed by #272

PR #272, opened 2017-11-30
Removed createRights

brandonopened:
Changed createRights to editorRights until/if we figure out why it was that way in the first place

---
This change is [Reviewable](https://reviewable.io/reviews/opensalt/opensalt/272)


roverwolf:
This fixes #270

PR #291, opened 2017-12-11
Feature/validation

JosephLeon:
Add code to update children on import, log those children that are updated, log new items added to the doc.

---
This change is [Reviewable](https://reviewable.io/reviews/opensalt/opensalt/291)


app/Resources/assets/js/lsdoc/index.js
```diff
@@ -252,7 +252,7 @@ var Import = (function() {
         };
 
         $.ajax({
-            url: '/cf/github/import',
+            url: '/app_dev.php/cf/github/import',
```
lmsilvera: you should remove `app_dev.php/`

PR #299, opened 2017-12-14
Fixed UI Issues (288,296,)

JosephLeon:
Created the PR for Anuradha

---
This change is [Reviewable](https://reviewable.io/reviews/opensalt/opensalt/299)


roverwolf:
@anuradhas248 Trying the branch directly, it looks like if the user is not logged in then the right detail section is pushed down below the tree instead of beside it.

This is on a regular browser (chrome).

But when the user is logged in then it is looks fine.

selection_20171214t141630

anuradhas248:
@roverwolf  - I have fixed mentioned issue. Also resolved the conflict. Please verify and confirm.

JorgeLu1s:
@anuradhas248 you still have the bug when the user is not logged in, the right section is pushed down below the tree. (tested on firefox)

screenshot-2017-12-15 act holistic framework - mathematics

JorgeLu1s:
@anuradhas248 I noticed some stranger behaviour when you resize the page, the right section is pushed below the tree (like the previous bug when the user wasn't logged in) I compared against staging. Not sure if this is exactly a bug or if this is the expected behaviour.
Local:

local

Staging:

staging

anuradhas248:
@roverwolf - I have made tree structure responsive after 768 px size. And this is standard size. So for below 768px width(Resizing window) you will find this behavior. Even live site also behaving the same. As I didn't added code for responsive functionality on staging, you found different behavior. Please refer production site for the same.

roverwolf:
@anuradhas248 Note that the previous comment was from @JorgeLu1s not me.

Also note that your test case for the toggles now fails.

vinisheth07:
@roverwolf - I have fixed test cases. Please check and confirm.

roverwolf:
@JorgeLu1s Please review and approve if you are satisfied with the PR.

app/Resources/assets/js/cftree/view-modes.js
```diff
@@ -392,6 +392,16 @@ apx.chooserMode.initialize = function() {
     $("#tree1Instructions").hide();
     $("#treeRightSideMode").hide();
     $("#itemOptionsWrapper").hide();
+    
+    // PW: Added 12/8/2017 to compensate for updated "container toggles"
+    $("#treeSideLeft").width("100%");
+    $(".toggle-container-left").hide();
+    $(".treeSideLeftInner").height("100%");
+    $(".toggle-container-right").hide();
+    // $(".treeSideRightInner").height("100%");
```
JorgeLu1s: you should remove this, is commented anyway.

app/Resources/assets/sass/cftree/view.scss
```diff
@@ -1,19 +1,22 @@
...
+    padding: 8px;
+}
+body{
+    overflow-x: hidden;
 }
-
 #treeSideLeft {
-    border-right:3px solid #f5f5f5;
-    margin-right:-3px;
+    // border-right:3px solid #f5f5f5;
```
JorgeLu1s: you should remove this, is commented anyway

app/Resources/assets/sass/cftree/view.scss
```diff
@@ -1,19 +1,22 @@
...
+    float:left;
+    width:50%;
+    min-width: 300px;
 }
 
 #treeSideRight {
-    border-left:3px solid #f5f5f5;
+    //  border-left:3px solid #f5f5f5;
```
JorgeLu1s: you should remove this, is commented anyway.

app/Resources/assets/sass/cftree/view.scss
```diff
@@ -93,9 +130,11 @@ ul.fancytree-container {
 }
 
 span.fancytree-title {
-    max-width:90%;
-    overflow:hidden;
-    text-overflow:ellipsis;
+    //white-space: pre-wrap;
```
JorgeLu1s: you should remove this, is commented anyway.

app/Resources/assets/sass/cftree/view.scss
```diff
@@ -93,9 +130,11 @@ ul.fancytree-container {
...
+    //word-break: normal;
```
JorgeLu1s: same here

app/Resources/assets/sass/cftree/view.scss
```diff
@@ -442,6 +480,9 @@ div.fancytree-drag-helper span.fancytree-childcounter {
     }
 }
 
+//span.fancytree-node{
```
JorgeLu1s: you should remove this block, is commented anyway

roverwolf:
Reviewed 2 of 2 files at r2, 2 of 2 files at r3, 1 of 1 files at r4, 3 of 3 files at r5, 2 of 2 files at r6.
Review status: all files reviewed at latest revision, 7 unresolved discussions.

---

*[tests/acceptance/DraggerCest.php, line 39 at r6](https://reviewable.io:443/reviews/opensalt/opensalt/299#-L0lBH8-0JPsJswTcuAv:-L0lBH80kwYWM1R5ynh4:bsavd8q) ([raw file](https://github.com/opensalt/opensalt/blob/19efb0c97a2d9832e22b8e00e5601fe36c3a3c47/tests/acceptance/DraggerCest.php#L39)):*
> ```PHP
>         $I->click('#toggleRight');
>         $left = $I->executeJS('return $("#treeSideLeft").width()');
>         $I->assertEquals($left,'1471');
> ```

Using a hardcoded value for the size fails the test when a different browser is used to run the test (chrome vs firefox).

It would be better to do the comparison relatively to the element's parent for example (and probably check that both the left and right did what was expected).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/opensalt/opensalt/299)*

roverwolf:
Reviewed 1 of 1 files at r7.
Review status: all files reviewed at latest revision, 6 unresolved discussions.

---

*[tests/acceptance/DraggerCest.php, line 39 at r6](https://reviewable.io:443/reviews/opensalt/opensalt/299#-L0lBH8-0JPsJswTcuAv:-L0lKg_4bZBqqyEzUZDm:baqzxql) ([raw file](https://github.com/opensalt/opensalt/blob/19efb0c97a2d9832e22b8e00e5601fe36c3a3c47/tests/acceptance/DraggerCest.php#L39)):*

Using a hardcoded value for the size fails the test when a different browser is used to run the test (chrome vs firefox).

It would be better to do the comparison relatively to the element's parent for example (and probably check that both the left and right did what was expected).
</blockquote></details>

I have went ahead and updated the test.  Now passes in chrome as well as firefox.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/opensalt/opensalt/299)*

Issue #242, opened 2017-11-09
Markdown editor full screen shoudl use the entire screen


jmarks:
When editing an item in the markdown editor, full screen is constrained to the the size of the dialog box for edit item.

brandonopened:
To be clear, hitting full screen for the editor should make the entire dialog box full screen? Why not just make the dialog box full screen to begin with? 
<img width="1152" alt="screen shot 2017-11-09 at 1 38 31 pm" src="https://user-images.githubusercontent.com/14943151/32630945-5f2397de-c553-11e7-9d27-bbeecb607d41.png">
<img width="1158" alt="screen shot 2017-11-09 at 1 38 00 pm" src="https://user-images.githubusercontent.com/14943151/32630946-5f445118-c553-11e7-9fe3-35ae70d141a2.png">

jmarks:
No it should make the body or note field being edited full screen, not the whole item record.

anuradhas248:
Do you want full monitor screen occupied or only modal screen (Form screen ) ?

roverwolf:
Should the editing (or adding) of an item even be in a dialogue box?  Or should it just be put in the page directly?

jmarks:
Editing an item or document in the entire page would be an improvement over using a dialog box for a form, 

None the less, dialog or not, in the MarkDown editor for the item full statement or notes filed, when that is select to go full screen only the field editing interface should take the full screen, and the rest of the form is not available until you leave full screen mode for the form element.

brandonopened:
Perhaps once we get #259 implemented, we could "force full screen toggle" of the editing screen to achieve the goals of this ticket.

BarbAllen:
Jira #2706

PR #302, opened 2017-12-19
Edit screen force to display full screen

anuradhas248:
Created the PR for Nilesh- Made all edit screen pop ups to full screen.  Ticket ids=KMS-2867, KMS-2938

---
This change is [Reviewable](https://reviewable.io/reviews/opensalt/opensalt/302)


roverwolf:
@anuradhas248 What are we trying to accomplish here?
Can you create GitHub issues for the referenced tickets, or at least add the content of those tickets in  pull request comments or the description?

I will note that some of the modals look very odd:
 - is almost the full width of the window, but flush to the left and leaves some space on the right
 - the width of the modal does not take into account the size of the screen -- on wide windows the main page is not as wide as the screen but the modal is
 - shorter modals such as delete and add exemplar look out of place
 - narrow modals such as delete have a lot of white space between the text on the left and the action buttons on the right

brandonopened:
came from #242 .

NileshPatreLM:
@roverwolf 
Please ind my comments in bold
1. is almost the full width of the window, but flush to the left and leaves some space on the right - **please let me know how it looks now**

 2. the width of the modal does not take into account the size of the screen -- on wide windows the main page is not as wide as the screen but the modal is - **this is because model pop ups are responsive and opensalt landing page header does not adjust properly according to screen size.**

 3.shorter modals such as delete and add exemplar look out of place - **fixed**
 4. narrow modals such as delete have a lot of white space between the text on the left and the action buttons on the right-**fixed**

roverwolf:
It looks better now, still seem to be a bit offset to the left, but I think that is due to other issues introduced in with the previous change to add the ability to adjust the width between the sides.  (Documented in #303 )

---

Reviewed 3 of 4 files at r2.
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---

*[features/FullScreenModalCest.php, line 9 at r2](https://reviewable.io:443/reviews/opensalt/opensalt/302#-L22q1FYLI3TQqij9Ip6:-L22q1FYLI3TQqij9Ip7:b-pstduf) ([raw file](https://github.com/opensalt/opensalt/blob/5039d5537467fd653733752ec982dc41bb4f6150/features/FullScreenModalCest.php#L9)):*
> ····

The "blank" line on line 9 and line 22 have 4 spaces instead of being empty.

---

*[src/CftfBundle/Resources/views/Editor/_fullScreenModalComponent.html.twig, line 1 at r2](https://reviewable.io:443/reviews/opensalt/opensalt/302#-L22qqytjvJyNRKFb2FA:-L22qqytjvJyNRKFb2FB:bw81cpp) ([raw file](https://github.com/opensalt/opensalt/blob/5039d5537467fd653733752ec982dc41bb4f6150/src/CftfBundle/Resources/views/Editor/_fullScreenModalComponent.html.twig#L1)):*
> ```Twig
> {% if closeButton is not defined %}
> ```

The `_fullScreenModalComponent.html.twig` file should probably extend the `_modalComponent.html.twig` file and add an empty block in the base modal file to be replaced with the style information instead of duplicating the entire thing.

https://twig.symfony.com/doc/2.x/tags/extends.html 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/opensalt/opensalt/302#-:-L22pcHCjnL0hGyF8EIW:bxl5jjk)*

PR #304, opened 2017-12-21
Export comments in csv format

anuradhas248:
Generate comment report in csv format. Story id: KMS-2256.

---
This change is [Reviewable](https://reviewable.io/reviews/opensalt/opensalt/304)


app/Resources/assets/js/lsdoc/index.js
```diff
@@ -649,3 +649,10 @@ $(window).resize(function(){
     $('#treeSideLeft').height($("#treeView").height());
 });
 
+
+$(".export_csv_comment").click(function(){
+    var currentUrl=window.location.href;
+    var urlArray = currentUrl.split('/');
+    var url="/salt/case/export_comment/"+urlArray[urlArray.length-2]+"/"+urlArray[urlArray.length-1]+"/comment.csv";
```
JorgeLu1s: You should be able to get the item id from the view by exposing it to a JavaScript variable, you don't need to ask the URL directly. Getting it from the URL itself could lead to troubles in the future if the URL format changes.

app/Resources/assets/js/lsdoc/index.js
```diff
@@ -649,3 +649,10 @@ $(window).resize(function(){
     $('#treeSideLeft').height($("#treeView").height());
 });
 
+
+$(".export_csv_comment").click(function(){
```
JorgeLu1s: This block should be a function in the comments.js file

src/CftfBundle/Resources/views/DocTree/view.html.twig
```diff
@@ -247,8 +253,14 @@
                                                         data-target="#seeImportLogs" data-alt-document-disabled="true">
                                                     Import Logs
                                                 </button>
-                                            {% endif %}
+                                            {% endif %}                                            
                                         {% endif %}
+                                        {% if 'comments' is active feature %}
```
JorgeLu1s: Is there a reason that this `if` block is repeated? (line 202 and 258)

src/Salt/SiteBundle/Controller/CommentsController.php
```diff
@@ -166,4 +167,60 @@ private function apiResponse($data, $statusCode = 200): JsonResponse
...
+     * 
+     * @param string $itemType
+     * 
+     * @param Request $request
+     *
+     * @return Response
+     */
+    public function exportCommentAction(string $itemType, int $id,Request $request)
+    {        
+        $url=$request->getBaseUrl('_route').'/cftree/'.$itemType.'/'.$id;
```
JorgeLu1s: Is the URL column mandatory? can't we just export the item type and the item id?.
We would need to refactor if the URL changes the format.

src/Salt/SiteBundle/Entity/CommentRepository.php
```diff
@@ -76,4 +84,42 @@ public function findByTypeItem(array $id): array
...
+     /**
+     * @param string $itemType
+     * @param int $itemId
+     *
+     * @return Comment
+     */    
+    
+    public function exportComments($itemType,$itemId)
+    {
+        $query= $this->getEntityManager()
```
JorgeLu1s: Is there a special reason to create manually the query instead of just using Doctrine ORM?

src/Salt/SiteBundle/Entity/CommentRepository.php
```diff
@@ -76,4 +84,42 @@ public function findByTypeItem(array $id): array
...
+        if($itemType=='item')
+        {
+            $query->select('r,u,o,i');
+        }
+        else 
+        {
+           $query->select('r,u,o,d');
+        }
+        $query->from('SaltSiteBundle:Comment', 'r') 
+              ->innerJoin('r.user','u')
```
JorgeLu1s: Is there a reason to add this join? why would you need to do this? you only need the item id

src/Salt/SiteBundle/Entity/CommentRepository.php
```diff
@@ -76,4 +84,42 @@ public function findByTypeItem(array $id): array
...
+              ->innerJoin('u.org','o');
```
JorgeLu1s: Is there a reason to add this join? why would you need to do this? you only need the item id

JorgeLu1s:
@anuradhas248 please note that you should refactor the unit tests since you refactored the Comment entity.

src/CftfBundle/Resources/views/DocTree/view.html.twig
anuradhas248: One block is for document and other one is for item div. Both div can't see at a time.

src/Salt/SiteBundle/Controller/CommentsController.php
anuradhas248: Yes, In story acceptance criteria mentioned that column. As write now we are not storing url in db, we have to create manually.

src/Salt/SiteBundle/Entity/CommentRepository.php
anuradhas248: I want to fit everything in single query. So I preferred this way. If you have other better way which fulfills all requirements in single query, then would you mind suggesting that query.

src/Salt/SiteBundle/Entity/CommentRepository.php
anuradhas248: In story acceptance criteria mentioned user name column too. And user name can get only by joining tables.

src/Salt/SiteBundle/Entity/CommentRepository.php
anuradhas248: In story acceptance criteria mentioned organization name column too. And organization name can get only by joining tables.

app/Resources/assets/js/lsdoc/index.js
anuradhas248: Yes, We might be get item id by you're suggested way. We discussed this thing in one of hotline and we were not sure how to get modified id by that way. Even you were fine with current used way that time. If you still think we should need to use that way only, then I need some time to research on it.

anuradhas248:
@JorgeLu1s :  All comment test cases are in skipped state. So I am not sure, which test case is falling or need to refactor.

app/Resources/assets/js/lsdoc/index.js
anuradhas248: Moved code to comment.js file

app/Resources/assets/js/lsdoc/index.js
JorgeLu1s: We need to refactor this. As I said in the hotline the controller is returning the `itemId` and the `documentId` so you should be able to get this values in the view.
Another way is getting the id from the node in the same way the comment is created, please check `comment.js` file to see how the comment is created.

src/Salt/SiteBundle/Entity/CommentRepository.php
JorgeLu1s: You can get all data you need with one single query with doctrine ORM, finding by `itemId`.

src/Salt/SiteBundle/Entity/CommentRepository.php
JorgeLu1s: The Comment entity has a relationship with the username, you should be able to get the User object from the Comment object (`$comment->getUser();`), so you don't need to add the join.

src/Salt/SiteBundle/Entity/CommentRepository.php
JorgeLu1s: As I said in the previous comment, the Comment entity has a relationship with the User and the User has a relationship with the Organization, so you should be able to get the Organization object from the User object ($user->getOrg();`)

JorgeLu1s:
@anuradhas248 almost all the unit tests need to be refactored since you modified the entity and the table in the database.
The unit tests shouldn't be skipped, please make sure you've enabled the comments feature before running the tests.

app/Resources/assets/js/lsdoc/comments.js
```diff
@@ -132,3 +132,10 @@ global.CommentSystem = CommentSystem;
 $(document).on('ready', function(){
     CommentSystem.init();
 });
+
+$(".export_csv_comment").click(function(){
```
JorgeLu1s: this block should be a function inside the `CommentSystem` var.

src/Salt/SiteBundle/Controller/CommentsController.php
```diff
@@ -183,21 +183,21 @@ public function exportCommentAction(string $itemType, int $id,Request $request)
     {        
         $url=$request->getBaseUrl('_route').'/cftree/'.$itemType.'/'.$id;
         $repo = $this->getDoctrine()->getManager()->getRepository(Comment::class);
-        $comment_data=$repo->exportComments($itemType,$id);
         $rows=array();
         if($itemType=='item')
```
JorgeLu1s: You can refactor this to skip the repeated block, also you can ask for `$itemType` and remove the if condition.

src/Salt/SiteBundle/Entity/Comment.php
```diff
@@ -35,7 +37,7 @@ class Comment
      * @var Comment
      *
      * @ORM\ManyToOne(targetEntity="\Salt\SiteBundle\Entity\Comment")
-     * @ORM\JoinColumn(nullable=false, onDelete="CASCADE")
+     * @ORM\JoinColumn(nullable=true, onDelete="CASCADE")
```
JorgeLu1s: Is there a special reason to make the user nullable? Also notice that you didn't add the migration to make it nullable.

src/Salt/SiteBundle/Controller/CommentsController.php
```diff
@@ -166,4 +167,59 @@ private function apiResponse($data, $statusCode = 200): JsonResponse
...
+    public function exportCommentAction(string $itemType, int $id, Request $request)
+    {
+        $url=$request->getBaseUrl('_route').'/cftree/'.$itemType.'/'.$id;
+        $repo = $this->getDoctrine()->getManager()->getRepository(Comment::class);
+        $rows=array();
+        if($itemType=='item')
+        {
+            $comments=array('Framework Name', 'Node Address', 'HumanCodingScheme', 'User', 'Organization', 'Comment');
+            $rows[]=implode(',', $comments);
+            $comment_data=$repo->findBy(['item'=>$id]);
```
JorgeLu1s: You can pass `$itemType` to get the comments data and remove the `if` block condition

src/Salt/SiteBundle/Controller/CommentsController.php
```diff
@@ -166,4 +167,59 @@ private function apiResponse($data, $statusCode = 200): JsonResponse
...
+                    );
+                  $rows[]=implode(',', $comments);
+            }
+        }
+        else
+        {
+            $comments=array('Framework Name', 'Node Address', 'User', 'Organization', 'Comment');
+            $rows[]=implode(',', $comments);
+            $comment_data=$repo->findBy(['document'=>$id]);
+            foreach($comment_data as $comment)
```
JorgeLu1s: We need to refactor to remove this repeated block

tests/unit/CommentTest.php
```diff
@@ -13,21 +15,23 @@ class CommentTest extends \Codeception\Test\Unit
 
     // tests
     public function testAddComment()
```
JorgeLu1s: We would need to test the same action with `document` too

jmarks:
Please make sure to test the CSV export with the comments modules configured off. Similarly, please test and handle import of CSV with comments into a server with comments configured off.

roverwolf:
@jmarks Note that **this** PR is about exporting a CSV of the comments.  So yes, it is to test that the button does not appear when the comment feature is off.  But there is no import involved here.

src/Salt/SiteBundle/Controller/CommentsController.php
```diff
@@ -146,80 +149,64 @@ public function downvoteAction(Comment $comment, UserInterface $user)
...
         return $this->get('jms_serializer')
-            ->serialize($data, 'json');
+        ->serialize($data, 'json');
     }
 
-    private function apiResponse($data, $statusCode = 200): JsonResponse
+    private function apiResponse($data, $statusCode=200): JsonResponse
     {
-        $json = $this->serialize($data);
+    $json=$this->serialize($data);
```
JorgeLu1s: We should fix the wrong indentation in this block

src/Salt/SiteBundle/Controller/CommentsController.php
```diff
@@ -146,80 +149,64 @@ public function downvoteAction(Comment $comment, UserInterface $user)
...
-     * @Route("/salt/case/export_comment/{itemType}/{id}/comment.csv", name="export_comment_file")
-     *
-     * @param int $id
-     * @param string $itemType
-     * @param Request $request
-     *
-     * @return Response
-     */
-    public function exportCommentAction(string $itemType, int $id, Request $request)
+/**
```
JorgeLu1s: Fix wrong indentation in this entire block. Also we should keep private functions at the end of the file.

src/Salt/SiteBundle/Controller/CommentsController.php
```diff
@@ -146,80 +149,64 @@ public function downvoteAction(Comment $comment, UserInterface $user)
...
+        $comment=$em->getRepository('SaltSiteBundle:Comment')->addComment($itemType, $itemId, $user, $content, $parentId);
 
         return $this->apiResponse($comment);
     }
 
     private function serialize($data)
     {
         return $this->get('jms_serializer')
-            ->serialize($data, 'json');
+        ->serialize($data, 'json');
```
JorgeLu1s: we shouldn't remove this indentation

src/Salt/SiteBundle/Controller/CommentsController.php
```diff
@@ -77,7 +79,7 @@ public function listAction(array $comments, UserInterface $user = null)
      */
     public function updateAction(Comment $comment, Request $request, UserInterface $user)
     {
-        $em = $this->getDoctrine()->getManager();
+        $em=$this->getDoctrine()->getManager();
```
JorgeLu1s: we shouldn't remove this space

src/Salt/SiteBundle/Controller/CommentsController.php
```diff
@@ -57,10 +57,12 @@ public function newItemCommentAction(Request $request, LsItem $item, UserInterfa
      *
      * @return mixed
      */
-    public function listAction(array $comments, UserInterface $user = null)
+    public function listAction(array $comments, UserInterface $user=null)
```
JorgeLu1s: we shouldn't remove this space

src/Salt/SiteBundle/Controller/CommentsController.php
```diff
@@ -94,7 +96,7 @@ public function updateAction(Comment $comment, Request $request, UserInterface $
      */
     public function deleteAction(Comment $comment, UserInterface $user)
     {
-        $em = $this->getDoctrine()->getManager();
+        $em=$this->getDoctrine()->getManager();
```
JorgeLu1s: we shouldn't remove this space

src/Salt/SiteBundle/Controller/CommentsController.php
```diff
@@ -22,6 +22,7 @@
 use Sensio\Bundle\FrameworkExtraBundle\Configuration\Security;
 use Salt\SiteBundle\Entity\Comment;
 use Qandidate\Bundle\ToggleBundle\Annotations\Toggle;
+use Symfony\Component\HttpFoundation\Response;
```
roverwolf: Line 25 is a duplicate of line 19 and should be removed.

src/Salt/SiteBundle/Controller/CommentsController.php
JosephLeon: This should be fixed.

JorgeLu1s:
Reviewed 1 of 2 files at r3, 1 of 3 files at r6, 2 of 4 files at r7, 7 of 7 files at r10.
Review status: all files reviewed at latest revision, 18 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/opensalt/opensalt/304)*

roverwolf:
Reviewed 1 of 2 files at r3, 1 of 3 files at r6, 1 of 4 files at r7, 2 of 7 files at r10.
Review status: all files reviewed at latest revision, 8 unresolved discussions.

---

*[src/CftfBundle/Resources/views/DocTree/view.html.twig, line 226 at r9](https://reviewable.io:443/reviews/opensalt/opensalt/304#-L22e3zhLv912rg7nigB:-L22e3zhLv912rg7nigC:bcky4f4) ([raw file](https://github.com/opensalt/opensalt/blob/44ae61bb3634598f55d2505b0d6471b7cc01e9ce/src/CftfBundle/Resources/views/DocTree/view.html.twig#L226)):*
> ```Twig
>                                         {% if manageEditorsRights %}
>                                             <a href="{{ path('framework_acl_edit', {'id':lsDocId}) }}"
>                                                class="btn btn-default btn-sm" data-alt-document-disabled="true">{{ lsItemId }} Manage
> ```

Why are we displaying the `lsItemId` here?

---

*[src/Salt/SiteBundle/Controller/CommentsController.php, line 184 at r1](https://reviewable.io:443/reviews/opensalt/opensalt/304#-L0t_6sU4ikIJ989U7rD-r1-184:-L22fMG4fStA7c-L6zui:bvjp4qp) ([raw file](https://github.com/opensalt/opensalt/blob/26f82db6b820c9ff9f8d1d73d94a04b93282f227/src/Salt/SiteBundle/Controller/CommentsController.php#L184)):*
<details><summary><i>Previously, anuradhas248 (Anuradha Suryawanshi) wrote…</i></summary><blockquote>

Yes, In story acceptance criteria mentioned that column. As right now we are not storing url in db, we have to create manually.
</blockquote></details>

This should probably use `$this->generateUrl(...)` instead of trying to manually generate the URL.

---

*[src/Salt/SiteBundle/Controller/CommentsController.php, line 164 at r10](https://reviewable.io:443/reviews/opensalt/opensalt/304#-L22suwJPgyIB-ZopeNo:-L22suwJPgyIB-ZopeNp:btrm6n5) ([raw file](https://github.com/opensalt/opensalt/blob/33a2a0f1e690c3a48f071fcd1c48b1e89b28a450/src/Salt/SiteBundle/Controller/CommentsController.php#L164)):*
> ```PHP
>     public function exportCommentAction(string $itemType, int $id, Request $request)
>     {
>         $url = $request->getBaseUrl('_route').'/cftree/'.$itemType.'/'.$id;
> ```

This should probably use `$this->generateUrl(...)` instead of hard-coding the url here.

---

*[src/Salt/SiteBundle/Controller/CommentsController.php, line 168 at r10](https://reviewable.io:443/reviews/opensalt/opensalt/304#-L22tp_Qoiqa7E_iIcKP:-L22tp_Qoiqa7E_iIcKQ:b-sn6urg) ([raw file](https://github.com/opensalt/opensalt/blob/33a2a0f1e690c3a48f071fcd1c48b1e89b28a450/src/Salt/SiteBundle/Controller/CommentsController.php#L168)):*
> ```PHP
>         $rows = array();
> 
>         $headers = array('Framework Name', 'Node Address', ($itemType == 'item') ? 'HumanCodingScheme' : null, 'User', 'Organization', 'Comment');
> ```

Should use === here instead of ==.

Should use [] syntax over array()

---

*[src/Salt/SiteBundle/Controller/CommentsController.php, line 200 at r10](https://reviewable.io:443/reviews/opensalt/opensalt/304#-L22uODMJwfR1EggcS5j:-L22uODMJwfR1EggcS5k:b23bi8u) ([raw file](https://github.com/opensalt/opensalt/blob/33a2a0f1e690c3a48f071fcd1c48b1e89b28a450/src/Salt/SiteBundle/Controller/CommentsController.php#L200)):*
> ```PHP
>      * @return JsonResponse
>      */
>     private function addComment(Request $request, string $itemType, $itemId, UserInterface $user): Response
> ```

The name of `$itemId` should probably be changed now that it is not an id.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/opensalt/opensalt/304)*

anuradhas248:
@roverwolf - I have made changes as per your comments. Please verify from your side.

roverwolf:
Reviewed 1 of 4 files at r7, 4 of 7 files at r10, 3 of 3 files at r11.
Review status: all files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/opensalt/opensalt/304)*

src/Salt/SiteBundle/Entity/CommentRepository.php
```diff
@@ -26,7 +26,14 @@ public function addComment($itemType, $itemId, User $user, string $content, $par
         $comment = new Comment();
```
roverwolf: The annotation still shows that `$itemId` is an integer (as does the name), but it is not.

src/Salt/SiteBundle/Controller/CommentsController.php
roverwolf: Line 25 is still there as a duplicate of line 19 and needs to be removed.

PR #308, opened 2018-01-05
Added email and user_type columns to users table

JorgeLu1s:
related to KMS-2458

---
This change is [Reviewable](https://reviewable.io/reviews/opensalt/opensalt/308)

### DIFF
--- a/app/DoctrineMigrations/Version20180105225202.php
+++ b/app/DoctrineMigrations/Version20180105225202.php
@@ -1,0 +1,30 @@
+<?php declare(strict_types = 1);
+
+namespace Application\Migrations;
+
+use Doctrine\DBAL\Migrations\AbstractMigration;
+use Doctrine\DBAL\Schema\Schema;
+
+/**
+ * Auto-generated Migration: Please modify to your needs!
+ */
+class Version20180105225202 extends AbstractMigration
+{
+    public function up(Schema $schema)
+    {
+        // this up() migration is auto-generated, please modify it to your needs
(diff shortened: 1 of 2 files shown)


brandonopened:
I understand the desire to not accept this until there is a use case for it; the idea was to continue to do these types of PR's in stages. IE the DB work, then the UI work it relies on? I'm fine either way just adding color as to why it was done this way.

roverwolf:
Without being linked to an issue (even as a part of the issue) and not even a description of how the fields are to be used it is impossible to review PRs like this.

What are the fields for?  Is the *email* field intended to replace the *username* field in some way?  Or is it a record of if the user has allowed emails to be sent to them?  How does the *user_type* field differ from the existing *roles*?  etc.  They are both marked as required fields but no data exists for them, so they cannot be applied as-is anyway.

brandonopened:
#371 Add validation to account_self_creation page to make username an email address

roverwolf:
@brandonopened Note that #371 might actually remove the need for this PR.
- [ ] #371 has the username being an email address, this PR adds it as a separate field.
- [ ] This PR also adds a "user type" column that has no definition for what it is for (255 character string).

brandonopened:
Jorge what is usertype?

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

brandonopened:
Very true. I think we should add email address a field for potential other uses(eventually auto-password reset, commenting); we can't force username to be an email address yet because like you said not all instances will require it.

roverwolf:
Closing as a new PR has been created that would obsolete this one.

PR #323, opened 2018-01-22
Pending status

JorgeLu1s:
Added pending status. Set "ROLE_VIEWER" as default role

### DIFF
--- a/app/DoctrineMigrations/Version20180115230815.php
+++ b/app/DoctrineMigrations/Version20180115230815.php
@@ -1,0 +1,33 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Application\Migrations;
+
+use Doctrine\DBAL\Migrations\AbstractMigration;
+use Doctrine\DBAL\Schema\Schema;
+
+/**
+ * Auto-generated Migration: Please modify to your needs!
+ */
+class Version20180115230815 extends AbstractMigration
+{
+    public function up(Schema $schema)
+    {
+        // this up() migration is auto-generated, please modify it to your needs
+        $this->abortIf($this->connection->getDatabasePlatform()->getName() !== 'mysql', 'Migration can only be executed safely on \'mysql\'.');
+
+        $this->addSql('ALTER TABLE salt_user ADD status INT DEFAULT 2 NOT NULL');
+        $this->addSql('UPDATE salt_user SET status = locked');
+        $this->addSql('ALTER TABLE salt_user DROP locked');
+    }
+
+    public function down(Schema $schema)
+    {
+        // this down() migration is auto-generated, please modify it to your needs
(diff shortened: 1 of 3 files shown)


app/DoctrineMigrations/Version20180115230815.php
```diff
@@ -0,0 +1,33 @@
...
+        $this->addSql('ALTER TABLE salt_user DROP locked');
+    }
+
+    public function down(Schema $schema)
+    {
+        // this down() migration is auto-generated, please modify it to your needs
+        $this->abortIf($this->connection->getDatabasePlatform()->getName() !== 'mysql', 'Migration can only be executed safely on \'mysql\'.');
+
+        $this->addSql('UPDATE salt_user SET locked = status');
+        $this->addSql('ALTER TABLE salt_user ADD locked TINYINT(1) DEFAULT \'0\' NOT NULL, DROP status');
```
roverwolf: Need to add the column back before the previous statement that uses it.

src/Salt/UserBundle/Entity/User.php
```diff
@@ -28,6 +28,10 @@ class User implements AdvancedUserInterface, \Serializable, EquatableInterface
         'ROLE_SUPER_USER',
     ];
 
+    private const ACTIVE = 0;
```
roverwolf: These should probably be **public** const values instead of private.

src/Salt/UserBundle/Entity/User.php
```diff
@@ -81,9 +85,9 @@ class User implements AdvancedUserInterface, \Serializable, EquatableInterface
     /**
      * @var bool
      *
-     * @ORM\Column(name="locked", type="boolean", nullable=false, options={"default": 0})
+     * @ORM\Column(name="status", type="integer", nullable=false, options={"default": 2})
```
roverwolf: The default should probably use the constant

src/Salt/UserBundle/Entity/User.php
```diff
@@ -435,9 +439,17 @@ public function suspendUser() {
      *
      * @return $this
      */
-    public function unsuspendUser() {
-        $this->locked = false;
+    public function activateUser() {
```
roverwolf: The admin UI needs to be updated to change the current "Unsuspend" action to have the new options.

tests/_support/Helper/UserManagement.php
```diff
@@ -110,7 +110,7 @@ protected function ensureUserExistsWithRoleLocal(string $role, $status): UserMan
             ->getQuery()
             ->getOneOrNullResult();
 
-        if ($user) {
+        if ($user && !is_null($status)) {
```
roverwolf: It might be better to try finding a user with the correct status in the query getting the user than to always create a new user when the status is null (should `null` even be allowed? or should it default to a specific status?).

PR #328, opened 2018-01-24
User List Filter 

anuradhas248:
Added filter on Organization and User Role column for User List page .
Story ID- KMS-3020

---
This change is [Reviewable](https://reviewable.io/reviews/opensalt/opensalt/328)

### DIFF
--- a/src/Salt/UserBundle/Form/Type/SearchForm.php
+++ b/src/Salt/UserBundle/Form/Type/SearchForm.php
@@ -12,7 +12,6 @@ use Symfony\Component\Form\FormBuilderInterface;
  */
 class SearchForm extends AbstractType
 {
-
     /**
      * {@inheritdoc}
      */
--- a/src/Salt/UserBundle/Form/Type/SearchForm.php
+++ b/src/Salt/UserBundle/Form/Type/SearchForm.php
@@ -20,17 +20,18 @@ class SearchForm extends AbstractType
         $builder
             ->add('organization', TextType::class, [
               'label' => 'Organization',
-                'required'   => false,
+              'required' => false,
             ])
             ->add('user_role', ChoiceType::class, [
-                'choices'  => [
+              'choices' => [
                 'Super User' => '["ROLE_SUPER_USER"]',
                 'Super Editor' => '["ROLE_SUPER_EDITOR"]',
                 'Admin' => '["ROLE_ADMIN"]',
                 'Editor' => '["ROLE_EDITOR"]',
                 'User' => '[]',
-                ],
-              'required'   => false,
-              ]);
+              ],
+              'required' => false,
+        ]);
     }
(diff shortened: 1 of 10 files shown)


tests/_support/Page/User.php
```diff
@@ -2,7 +2,6 @@
 
 namespace Page;
 
-
 use Behat\Behat\Context\Context;
 use Behat\Gherkin\Node\TableNode;
 use PhpSpec\Exception\Example\PendingException;
```
JosephLeon: Codacy is complaining about short variables here but we actually want to use `$I` as it makes sense when reading the code like gherkin style. The way to disable the codacy test here is to add this before the class:

```
/**
 * @SuppressWarnings(PHPMD.ShortVariable)
 */
class User implements Context {
```

src/Salt/UserBundle/Repository/OrganizationRepository.php
```diff
@@ -27,4 +27,20 @@ public function addNewOrganization($organizationName) {
...
+    /**
+     * Search all organizations which contains passed string.
+     *
+     * string|null $search
+     *
+     * @return array
+     */
+    public function findOrg($search): array
+    {
+        $qb = $this->createQueryBuilder('org');
```
JosephLeon: Codacy was complaining about the variable length, maybe replace `$qb` with `$query`.

src/Salt/UserBundle/Form/Type/SearchForm.php
```diff
@@ -0,0 +1,36 @@
...
+ */
+class SearchForm extends AbstractType
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function buildForm(FormBuilderInterface $builder, array $options)
+    {
+        $builder
+            ->add('organization', TextType::class, [
```
JosephLeon: Should be:

```
->add('organization', TextType::class, [
  'label' => 'Organization',
  'required' => false,
])
```

Try and make sure we don't have wrong indentation or extra spaces anywhere.

Also check lines 25-34 for the same issue.

anuradhas248:
@JosephLeon - I have taken care of your comments. Please verify and confirm.

JorgeLu1s:
@anuradhas248 Since you are implementing jquery datatables now, there is no need to keep the previous code, because it won't be used. I strongly suggest to close this PR, delete this branch and create a new branch with only the `$("#user_table").DataTables();` code, also would be better to add a `datatable` class to the html table and ask for that class in the javascript: `$('.datatable').DataTables();` so in the future we don't have to modify the javascript, just have add to add a class to the html table ;)

JorgeLu1s:
@anuradhas248 Also you have code related to the comments feature in this branch, you need to move this code to a different one and create a new PR for that.

PR #352, opened 2018-02-05
Add captcha

JorgeLu1s:
Added invisible captcha

---
This change is [Reviewable](https://reviewable.io/reviews/opensalt/opensalt/352)


brandonopened:
@JorgeLu1s ?

src/Salt/UserBundle/Resources/views/Security/login.html.twig
```diff
@@ -31,7 +35,11 @@
                         <input type="hidden" name="_target_path" value="{{ redirect }}"/>
                     {% endif %}
 
-                    <button class="pull-right btn btn-primary" type="submit">Login</button>
+                    {% if google_captcha_site_key is not empty %}
+                        <button class="pull-right btn btn-primary btn-login g-recaptcha" data-sitekey="{{ google_captcha_site_key }}" data-callback="onSubmit">Login</button>
+                    {% else %}
+                        <button class="pull-right btn btn-primary btn-login" data-callback="onSubmit">Login</button>
```
roverwolf: The regular submit button should probably remain a simple `type="submit"` button instead of having to call a javascript function (and wrap the js function below with the `if google...key ...` as well.

docker/.env.dist
```diff
@@ -42,3 +42,5 @@ FIREBASE_CLIENT_ID=
 FIREBASE_CLIENT_EMAIL=
 FIREBASE_PRIVATE_KEY=
 FIREBASE_PREFIX=opensalt
+GOOGLE_CAPTCHA_SITE_KEY=
+GOOGLE_CAPTCHA_SECRET_KEY=
```
roverwolf: The env variables need to be added to `docker/docker-compose.dev.yml` as well so that they will get passed into the container.

src/Salt/UserBundle/Listener/LoginListener.php
```diff
@@ -0,0 +1,33 @@
...
+
+    public function __construct(Container $container)
+    {
+        $this->container = $container;
+    }
+
+    public function onSecurityInteractiveLogin(InteractiveLoginEvent $event)
+    {
+        $request = $event->getRequest();
+        $captcha_secret = $this->container->getParameter('google_captcha_secret_key');
```
roverwolf: Would be better to just pass the parameter into the constructor instead of the entire container.

src/Salt/UserBundle/Listener/LoginListener.php
```diff
@@ -0,0 +1,34 @@
...
+namespace Salt\UserBundle\Listener;
+
+use Symfony\Component\DependencyInjection\ContainerInterface as Container;
+use Symfony\Component\Security\Core\Exception\BadCredentialsException;
+use Symfony\Component\Security\Http\Event\InteractiveLoginEvent;
+use ReCaptcha\ReCaptcha;
+
+class LoginListener
+{
+    private $container;
```
roverwolf: This should be `$captchaSecret` not `$container` to store the value being set.

src/Salt/UserBundle/Listener/LoginListener.php
```diff
@@ -0,0 +1,34 @@
+<?php
+
+namespace Salt\UserBundle\Listener;
+
+use Symfony\Component\DependencyInjection\ContainerInterface as Container;
```
roverwolf: The container use clause should be removed as is no longer required here.

PR #357, opened 2018-02-07
Add swiftmailer back into the app, add code to fire off an email during user creation

JosephLeon:
Swiftmailer was removed earlier in the development process since it wasn't being used but we added it back for requested functionality. Add code that utilizes swiftmailer to send an email during the user account creation process. Also added a plugin that helps codeception test that the email was actually sent.

---
This change is [Reviewable](https://reviewable.io/reviews/opensalt/opensalt/357)


src/Salt/UserBundle/Controller/UserController.php
```diff
@@ -81,6 +81,16 @@ public function newAction(Request $request)
...
 
+            // email the new user
+            $email = $targetUser->getUsername();
+            $message = (new \Swift_Message('Hello Email'))
+            ->setFrom('send@example.com')
+            ->setTo($email)
+            ->setSubject('Your account has been created')
+            ->setBody('Thank you! Your account has been created and you will be contacted in 2 business days when it is active.');
+
+            $this->get('mailer')->send($message);
```
roverwolf: The message should not be sent until after the command succeeds (in case creating the user fails).

We should not send an email if no mail service is configured.
Sending the email should be an optional feature that can be enabled/disabled.

tests/_support/Page/User.php
```diff
@@ -248,4 +248,19 @@ public function iApproveTheNewUser()
...
     }
 
+    /**
+     * @Then /^I verify an email was sent$/
+     */
+    public function IVerifyEmailWasSent()
+    {
+        $I = $this->I;
+
+        $I->fetchEmails();
```
roverwolf: There should probably be a check to see if an email server is configured and the feature turned on, and if not then skip the test.

tests/acceptance/features/user/super_user/add.feature
```diff
@@ -7,5 +7,5 @@ Feature: Add new User
   Scenario: 1016-1245 Adding new user
     Given I log in as a user with role "Super User"
     Then I add a new user with "Super User" role
-
-    Then I delete the User
\ No newline at end of file
+    Then I verify an email was sent
```
roverwolf: This should probably be a second test, so that creating a user can be tested separately from the email being sent (such as if the email sending is disabled).

app/config/config.yml
```diff
@@ -191,6 +197,20 @@ doctrine:
...
+    transport: "%mailer_transport%"
+    host:      "%mailer_host%"
+    username:  "%mailer_user%"
+    password:  "%mailer_password%"
+    # using an env var throws an error
+    # auth_mode: "%mailer_auth_mode%"
+    auth_mode: cram-md5
+    port:      "%mailer_port%"
+    # to disable mail uncomment MAILER_URL
+    # MAILER_URL: "null://localhost"
```
roverwolf: This does not appear true for Symfony 3.3/3.4 (there is a change in 4.0).

https://symfony.com/doc/3.4/email.html indicates a `disable_delivery` attribute can be used.

app/config/config.yml
```diff
@@ -191,6 +197,20 @@ doctrine:
...
+# Swiftmailer Configuration
+swiftmailer:
+    spool:     { type: memory }
+    transport: "%mailer_transport%"
+    host:      "%mailer_host%"
+    username:  "%mailer_user%"
+    password:  "%mailer_password%"
+    # using an env var throws an error
+    # auth_mode: "%mailer_auth_mode%"
+    auth_mode: cram-md5
```
roverwolf: Hmm.  Hard-coding this seems bad.  I wonder if there is another way to get the value set that will work.

src/Salt/UserBundle/Controller/UserController.php
JosephLeon: Moved the code after the user creation code in the try exception.

tests/acceptance/features/user/super_user/add.feature
JosephLeon: Moved the test into it's own email area in tests/acceptance/features/emails/super_user/user_create.feature

tests/acceptance/features/emails/super_user/user_create.feature
```diff
@@ -0,0 +1,11 @@
+Feature: Send email at user creation
+  In order to send an email at user creation
+  As a super user
+  I need to create a user
+
+  @super-user @user @add-user @1016-1245 @duplicate
```
roverwolf: - [ ] How will this pass since the email is no longer sent when a super-user adds a user?
- [ ] Should not be marked as a duplicate (what is it a duplicate of?)
- [ ] The @1016-1245 here and the next line should be changed to be a unique identifier for the test (current time in MMDD-hhmm format is usually unique, otherwise increase the "minute" value until unique)

app/config/config.yml
```diff
@@ -192,6 +198,21 @@ doctrine:
...
+swiftmailer:
+    spool:     { type: memory }
+    transport: "%mailer_transport%"
+    host:      "%mailer_host%"
+    username:  "%mailer_user%"
+    password:  "%mailer_password%"
+    # using an env var throws an error
+    # auth_mode: "%mailer_auth_mode%"
+    # auth_mode: "%env(MAILER_AUTH_MODE)%"
+    auth_mode: cram-md5
```
roverwolf: If we can figure it out, we should make this configurable via an environment variable.

app/config/config.yml
```diff
@@ -192,6 +198,21 @@ doctrine:
...
+    port:      "%mailer_port%"
+    # to disable mail uncomment disable_delivery: true
+    # disable_delivery: true
```
roverwolf: This should be configurable via an environment variable, direct changes to the `config.yml` file is not desired for deployments (only during development).

src/Salt/UserBundle/Controller/SignupController.php
```diff
@@ -66,14 +66,14 @@ public function signupAction(Request $request)
...
-                // $message = (new \Swift_Message('Hello Email'))
-                // ->setFrom('send@example.com')
-                // ->setTo($email)
-                // ->setSubject('Your account has been created')
-                // ->setBody('Thank you! Your account has been created and you will be contacted in 2 business days when it is active.');
-                //
-                // $this->get('mailer')->send($message);
+                $email = $targetUser->getUsername();
+                $message = (new \Swift_Message('Hello Email'))
+                ->setFrom('send@example.com')
```
roverwolf: The "From" email address needs to be configurable (so that it is valid).

src/Salt/UserBundle/Controller/SignupController.php
```diff
@@ -65,6 +65,16 @@ public function signupAction(Request $request)
                 $command = new AddUserCommand($targetUser, $encryptedPassword);
                 $this->sendCommand($command);
 
+                // email the new user
+                $email = $targetUser->getUsername();
+                $message = (new \Swift_Message('Hello Email'))
+                ->setFrom('send@example.com')
```
roverwolf: The from address needs to be configurable (so that it will be valid)

src/Salt/UserBundle/Controller/SignupController.php
```diff
@@ -67,6 +67,7 @@ public function signupAction(Request $request)
 
                 // email the new user
                 $email = $targetUser->getUsername();
+                
                 $message = (new \Swift_Message('Hello Email'))
                 ->setFrom('send@example.com')
```
roverwolf: The from address needs to be configurable (so that it will be valid)

app/config/config.yml
JosephLeon: I wasn't able to find a fix for this. We might want to make this an issue to solve later.

app/config/config.yml
JosephLeon: I wasn't able to find a fix for this. We might want to make this an issue to solve later.

tests/_support/Page/User.php
JosephLeon: I added an env var USE_MAIL_FEATURE so that we can control functionality throughout the app depending if mail functionality is wanted or not.

JosephLeon:
Resolves #357

JosephLeon:
Resolves #375

src/Salt/UserBundle/Controller/SignupController.php
roverwolf: The `from` email address still needs to be configurable.

docker/docker-compose.dev.yml
```diff
@@ -32,7 +32,14 @@ services:
...
       FIREBASE_PRIVATE_KEY:
       FIREBASE_PREFIX:
+      MAILER_TRANSPORT:
+      MAILER_HOST:
+      MAILER_USER:
+      MAILER_PASSWORD:
+      MAILER_AUTH_MODE:
+      MAILER_PORT:
       CREATE_ACCOUNT_FEATURE:
+      USE_MAIL_FEATURE:
```
roverwolf: Should this be configured as a "Feature Toggle" the same way the other features (commenting and create account) are?

src/Salt/UserBundle/Controller/SignupController.php
```diff
@@ -65,6 +65,16 @@ public function signupAction(Request $request)
                 $command = new AddUserCommand($targetUser, $encryptedPassword);
                 $this->sendCommand($command);
 
+                // email the new user
+                $email = $targetUser->getUsername();
+                $message = (new \Swift_Message('Hello Email'))
```
roverwolf: Should this be checking the `USE_MAIL_FEATURE` option being passed?

docker/docker-compose.dev.yml
JosephLeon: Added the config in the feature toggle config section

src/Salt/UserBundle/Controller/SignupController.php
JosephLeon: This is now checking for the email feature being active

src/Salt/UserBundle/Controller/SignupController.php
JosephLeon: This was always causing trouble during testing so left it hard coded for now

src/Salt/UserBundle/Controller/SignupController.php
```diff
@@ -65,6 +65,19 @@ public function signupAction(Request $request)
                 $command = new AddUserCommand($targetUser, $encryptedPassword);
                 $this->sendCommand($command);
 
+                // check to see if the env var is set to "true" mailer
+                if (getenv('USE_MAIL_FEATURE') == 'always-active') {
```
roverwolf: This should be checking if the feature is enabled, not looking at the environment variable directly.

Ref: [Toggle Bundle example usage](https://github.com/qandidate-labs/qandidate-toggle-bundle#example-usage)

src/Salt/UserBundle/Controller/SignupController.php
```diff
@@ -66,12 +66,13 @@ public function signupAction(Request $request)
                 $this->sendCommand($command);
 
                 // check to see if the env var is set to "true" mailer
-                if (getenv('USE_MAIL_FEATURE') == 'always-active') {
-                  $fromEmail = getenv('MAIL_FEATURE_FROM_EMAIL');
+                if ($this->getParameter('USE_MAIL_FEATURE') == 'always-active') {
```
roverwolf: This should be checking if the feature is enabled, not looking at the environment variable directly.

Ref: [Toggle Bundle example usage](https://github.com/qandidate-labs/qandidate-toggle-bundle#example-usage)

src/Salt/UserBundle/Controller/SignupController.php
```diff
@@ -65,6 +65,21 @@ public function signupAction(Request $request)
                 $command = new AddUserCommand($targetUser, $encryptedPassword);
                 $this->sendCommand($command);
 
+                // check to see if the env var is set to "true" mailer
+                if ($this->getParameter('USE_MAIL_FEATURE') == 'always-active') {
+                  // $fromEmail = getenv('MAIL_FEATURE_FROM_EMAIL');
+                  $fromEmail = $this->getParameter('MAIL_FEATURE_FROM_EMAIL');
```
roverwolf: The parameter should be the lower-case one (`mail_feature_from_email`) defined in `config.yml`

app/config/parameters.yml.dist
```diff
@@ -33,6 +35,9 @@ parameters:
     # to enable the comments feature you should pass 'always-active' instead of 'inactive'
     env(COMMENTS_FEATURE): inactive
     env(CREATE_ACCOUNT_FEATURE): inactive
+    env(USE_MAIL_FEATURE): inactive
+
+    env(MAIL_FEATURE_FROM_EMAIL): send@example.com
```
roverwolf: The default values should be usable.  `send@example.com` is not.
Probably should be set to ~ (representing null).

app/config/parameters.yml.no-env
```diff
@@ -27,6 +27,9 @@ parameters:
     # to enable the comments feature you should pass 'always-active' instead of 'inactive'
     env(COMMENTS_FEATURE): inactive
     env(CREATE_ACCOUNT_FEATURE): inactive
+    env(USE_MAIL_FEATURE): inactive
+
+    env(MAIL_FEATURE_FROM_EMAIL): send@example.com
```
roverwolf: The default values should be usable.  `send@example.com` is not.
Probably should be set to ~ (representing null).

app/config/parameters.yml.dist
roverwolf: (might also want to check for it being null before trying to send the email)

JosephLeon: Changed the value and also added conditional logic in both the controller and the test to make sure the from email value isn't null.

src/Salt/UserBundle/Controller/SignupController.php
JosephLeon: Fixed

PR #381, opened 2018-02-15
License on exported Excel file

### LINKED ISSUE #332: License and License URI fields for View a Framework - #283  KMS-3242
As an OpenSALT user, I am able to see the License title and License URI when viewing a Framework.
#283

lmsilvera:
Showing license text on exported CASE file.

closes #332

---
This change is [Reviewable](https://reviewable.io/reviews/opensalt/opensalt/381)

### DIFF
--- a/src/App/Service/ExcelExport.php
+++ b/src/App/Service/ExcelExport.php
@@ -98,12 +98,15 @@ class ExcelExport
     public function generateExcelFile(LsDoc $cfDoc, array $items, array $associations, array $smartLevel, \PHPExcel $phpExcelObject): void
     {
         $licenseTitle = "";
+        $licenseText = "";
         $phpExcelObject->getProperties()
             ->setCreator('OpenSALT')
             ->setTitle('case');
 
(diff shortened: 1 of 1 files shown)


src/App/Service/ExcelExport.php
```diff
@@ -130,6 +130,7 @@ public function generateExcelFile(LsDoc $cfDoc, array $items, array $association
             ->setCellValue('K2', $cfDoc->getAdoptionStatus())
             ->setCellValue('L2', $cfDoc->getStatusStart())
             ->setCellValue('M2', $cfDoc->getStatusEnd())
+            ->setCellValue('N2', $cfDoc->getLicence()->getLicenceText())
```
roverwolf: We might need to export more than just the text (which could be very long) if we want to be able to re-import the file.

roverwolf: If there is no licence set then `$cfDoc->getLicence()` will be `null`, so need to account for that here.

lmsilvera:
@roverwolf PR ready to be merged.

Issue #371, opened 2018-02-09
email address validation on create new user page


brandonopened:
AS A user without login access to OpenSALT
I WANT to be able to enter an email address as username when self-creating an account
SO THAT I can receive emails from the system about my account status. 

(validate username to be an email address and add text to the "Create New User" page reflecting that.

brandonopened:
So, this is something that will be reflected in the FRD. 

1) When the CREATE NEW USER page is turned on, the UI will validate that the username is an email address.

Issue #382, opened 2018-02-16
Error when let new org field blank on signup page


JorgeLu1s:
When creating a new account a 500 internal server error is shown if I let the 'new organization' field blank

JorgeLu1s:
closed in PR #386

PR #383, opened 2018-02-16
Add email to user, fix new organization blank error

JorgeLu1s:
fix issues #371 and #382

---
This change is [Reviewable](https://reviewable.io/reviews/opensalt/opensalt/383)

### DIFF
--- a/app/DoctrineMigrations/Version20180216003902.php
+++ b/app/DoctrineMigrations/Version20180216003902.php
@@ -1,0 +1,32 @@
+<?php declare(strict_types = 1);
+
+namespace Application\Migrations;
+
+use Doctrine\DBAL\Migrations\AbstractMigration;
+use Doctrine\DBAL\Schema\Schema;
+
+/**
+ * Auto-generated Migration: Please modify to your needs!
+ */
+class Version20180216003902 extends AbstractMigration
+{
+    public function up(Schema $schema)
+    {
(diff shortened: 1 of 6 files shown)


roverwolf:
#382 states that the username should be an email address, but this PR is adding an additional field to hold the email address.

Is there really a need for a new field to hold the email address instead of using the username?  It probably makes sense to just have the username as the email address.

If a new field is thought to be necessary, discussion should be added to #382.

roverwolf:
You might want to split this PR so that the organisation fix can be merged in right away while any discussion on the email field takes place.

JosephLeon:
I think we should use emails as usernames instead of having both usernames and emails. Will loop @brandonopened  in on this issue

brandonopened:
WIll making username an email address affect existing accounts?

roverwolf:
We can ensure it doesn't.
1. Make sure the email service does not try to send emails to a username that does not look like an email address
2. Allow (and possibly encourage?) users with non-email usernames to be changed to email addresses.

roverwolf:
Per #371 and Brandon's [message](https://opened.slack.com/archives/C6Y1RSQKC/p1518801240000188) on Slack we will **not** add an extra email field but instead have the username be the email address.

@JorgeLu1s Please update the PR to not add the new column.

JorgeLu1s:
ok @roverwolf I'm going to close this PR and create a new one just for the new organization fix

Issue #343, opened 2018-02-01
self service create new user email notification flow


brandonopened:
![email flow](https://user-images.githubusercontent.com/14943151/35687123-cdde316c-0765-11e8-88dd-37b1c250e92c.PNG)

roverwolf:
Assumptions:
 - The "self service create new user" option would only be enabled and available if there is an email server configured.

Questions:
 - The image makes it look like the user can create an organisation, I am assuming this is not the case and must choose from the current list.
 - What happens if the Admin deletes the user instead of aproving or suspending the user?

brandonopened:
correct, there will be an option to turn this on or off as on cfdoc there will be a button to 'create a new user'. 
https://opensalt.mybalsamiq.com/projects/opensalt/home
and
https://opensalt.mybalsamiq.com/projects/opensalt/Create%20an%20account-BA

- validation on the username not being a duplicate
- validation that the passwords are matching
- validation that the org name they enter is not a duplicate

Yes, we are saying a new org will be created with pending status as well
https://opensalt.mybalsamiq.com/projects/opensalt/Org%20Admin
(and same code for pending etc)

BUT the main 'action' is here on the user list page
https://opensalt.mybalsamiq.com/projects/opensalt/Org%20Admin

- When an admin approves a user their acccount becomes active. 
- If they had created a new org it will also be approved when the first user is approved

eventually we will send the notification to the org admin as selected on new user account creation, for now it will go to super user. (eg meta-admin).

jmarks:
Additional requirement: Please make this feature a admin configurable application option. We have users who will not want auto registration and will require only SSO authentication of manged account creation. This shoudl be considered along with SSO integration stories.

roverwolf:
#375 has been added as a first step to add the ability to do emails

PR #390, opened 2018-02-19
Send approval email when a super_user approves a user account 

JosephLeon:
When a public_user creates an account it will be in a "pending" state, then a super_user can go to the user list view and there will be an "Approve" button. When that button is clicked an email will be sent to the user notifying them of the accounts approval.

---
This change is [Reviewable](https://reviewable.io/reviews/opensalt/opensalt/390)


roverwolf:
ref: #343

PR #398, opened 2018-02-22
Signup validations

JorgeLu1s:
Add validations to sign up form.
Relagted to #371

---
This change is [Reviewable](https://reviewable.io/reviews/opensalt/opensalt/398)


src/Salt/UserBundle/Entity/User.php
```diff
@@ -66,6 +66,12 @@ class User implements AdvancedUserInterface, \Serializable, EquatableInterface
      *
      * @Assert\NotBlank(groups={"registration"})
      * @Assert\Length(max=4096)
+     * @Assert\Regex(
+     *     pattern="/(?=^.{8,}$)((?=.*\d)|(?=.*\W+))(?![.\n])(?=.*[A-Z])(?=.*[a-z]).*$/",
```
roverwolf: The regex does not appear to properly check for the rules listed.
- ex. 12345678a# fails but has {is least 8 chars long, a number, a letter, a special character}

Note that it will probably be difficult to write those rules as one RegEx due to the 3 of 4 part.  You might need to use a [custom validator](https://symfony.com/doc/3.4/validation/custom_constraint.html) (probably best), [expression](https://symfony.com/doc/current/reference/constraints/Expression.html), or [callback](https://symfony.com/doc/current/reference/constraints/Callback.html).

PR #394, opened 2018-02-20
Upgrade to PHPSpreadsheet

roverwolf:
PHPExcel has been deprecated and recommends upgrading to PHPSpreadsheet from the same project.

resolves #337

---
This change is [Reviewable](https://reviewable.io/reviews/opensalt/opensalt/394)

PR #401, opened 2018-02-22
License URI on exported Excel file

### LINKED ISSUE #400: License URI is missing on exported Excel file.
License URI is missing on exported Excel file.

lmsilvera:
License URI on exported Excel file

closes #400

---
This change is [Reviewable](https://reviewable.io/reviews/opensalt/opensalt/401)


roverwolf:
If the export is changed then the import should be updated as well to match.

Note the pending PR #394 to update the code to use PHPSpreadsheet intead of PHPExcel (@JorgeLu1s can you review that soon as it also has a fix to import the prior version of the export).

lmsilvera:
@roverwolf PR already updated and ready for review.

roverwolf:
@lmsilvera It does not look like the Excel import matches what is now being exported.

brandonopened:
@roverwolf yes I'm working on that right now. The export file looks different due to the more machine-readable changes on CF association but CfDoc and CfItem will be the same in a few minutes here including this change on the import template.

brandonopened:
Or Luis for now just add that extra line to the current importer so this PR can merge.

lmsilvera:
ok @brandonopened I'll do it.

src/App/Service/ExcelImport.php
```diff
@@ -129,16 +130,45 @@ private function saveDoc(Worksheet $sheet): LsDoc
...
 
+    private function getLicence(Worksheet $sheet): LsDefLicence
+    {
+        $uri = $this->getCellValueOrNull($sheet, 14, 2);
+        $title = $this->getCellValueOrNull($sheet, 15, 2);
+        $licenceText = $this->getCellValueOrNull($sheet, 16, 2);
+
+        // Look for this licence locally
+        $licence = $this->getEntityManager()->getRepository(LsDefLicence::class)
+            ->findOneByUri($uri);
```
roverwolf: Should probably check if `$uri` is empty before trying to find a licence that matches (or trying to create one), as it is possible for there to be no licence.

lmsilvera: sure thing.

lmsilvera:
@roverwolf PR ready

src/App/Service/ExcelImport.php
roverwolf: It would be better to check if `$uri` is empty before using it to query the database, but it will work.

PR #402, opened 2018-02-23
Fixed cache issue on child statement

anuradhas248:
Fixed cache issue on child statement.
So now Framework creator can see imported files -> bug KMS-3332 and KMS- 3211
and User can see updated data for child items -> bug KMS- 3493

---
This change is [Reviewable](https://reviewable.io/reviews/opensalt/opensalt/402)

### DIFF
--- a/app/Resources/views/base.html.twig
+++ b/app/Resources/views/base.html.twig
@@ -7,6 +7,7 @@
             <meta http-equiv="X-UA-Compatible" content="IE=edge,chrome=1">
         {% endblock %}
         <meta http-equiv="Content-Type" content="text/html; charset=utf-8">
+        <meta http-equiv="pragma" content="no-cache" />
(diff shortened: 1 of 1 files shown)


anuradhas248:
@roverwolf - I am not able to reproduce this issue on my local environment so I have checked this issue with fire bug on Staging. I found that when I disabled cache using fire bug and refreshed this issues is not reproducible.
So I have added pragma meta data in base.html.twig file and I am assuming this will fix issue on staging.
Kindly provide your comments on this.
Thank you

anuradhas248:
@JorgeLu1s , @JosephLeon - As per Dave's above comment this issue is already fixed in his PR. So we don't need this PR. Can one of you delete this PR?

roverwolf:
Closing unmerged per comment by @anuradhas248

PR #403, opened 2018-02-23
Add reject button (KMS-3615)

adlearningmate:
Behavior - clicking REJECT removes pending status and Suspend appears and Approve disappears
1) Login as Super User : New User create than display the show,edit,suspend, approve and reject click reject button display the show and unsuspend.
2) Login as  Organization Admin(Admin) : Display the show, edit,suspend and approve.

---
This change is [Reviewable](https://reviewable.io/reviews/opensalt/opensalt/403)


app/Resources/assets/sass/styles.scss
```diff
@@ -131,3 +131,9 @@ span.notification-url-indicator {
     margin-right: 15px;
     text-decoration: underline;
 }
+
+.btn-default.btn-reject{
+    color: #fff;
+    background-color: #f37321;
+
```
JosephLeon: Remove empty line

adlearningmate:
@JosephLeon 
Joe i have remove the empty line. Please check!

app/Resources/assets/sass/styles.scss
```diff
@@ -131,3 +131,8 @@ span.notification-url-indicator {
     margin-right: 15px;
     text-decoration: underline;
 }
+
+.btn-default.btn-reject{
```
JorgeLu1s: Bootstrap already have a `btn-danger` class to make the button red ( https://www.w3schools.com/bootstrap/bootstrap_buttons.asp for reference) so there is no need to manually create a new style, if you have an specific requirement to implement it this way, please ignore this comment. If not please use `btn-danger` instead. This way we can be consistent with the style guide

src/Salt/UserBundle/Resources/views/User/index.html.twig
```diff
@@ -43,26 +43,31 @@
...
+                                    {% endif %}
                                 </li>
-                            {% endif %}
+                                {% if user.pending %}
+                                    <li>
+                                        <a class="btn btn-primary" href="{{ path('admin_user_activate', { 'id': user.id }) }}">Approve</a>
+                                    </li>
+                                    {% if is_granted('ROLE_SUPER_USER') %} 
+                                        <li>
+                                            <a class="btn btn-default btn-reject" href="{{ path('admin_user_reject', { 'id': user.id }) }}">Reject</a>
```
JorgeLu1s: please read my previous comment about the `btn-danger`, would be better to have the same button style in all the views. As I said if the requirement is to use this color instead ignore this comment.

roverwolf: Also, there should not normally be a need to have both `btn-default` and another `btn-*` class on the same button.

src/Salt/UserBundle/Resources/views/user/index.html.twig
```diff
@@ -43,26 +43,31 @@
...
+                                    {% else %}
+                                        <a class="btn btn-default" href="{{ path('admin_user_suspend', { 'id': user.id }) }}">Suspend</a>
+                                    {% endif %}
                                 </li>
-                            {% endif %}
+                                {% if user.pending %}
+                                    <li>
+                                        <a class="btn btn-primary" href="{{ path('admin_user_activate', { 'id': user.id }) }}">Approve</a>
+                                    </li>
+                                    {% if is_granted('ROLE_SUPER_USER') %}
```
roverwolf: Does it make sense that only a user with the ROLE_SUPER_USER role can `Reject` a user but other can `Approve`, especially since others can also `Suspend` the user by the above logic (which does the same thing as `Reject` does)?

brandonopened:
The idea was that super user have 'approve' rights because that would allow them to approve new organizations too... if we just let org admin's approve then we would leave out the type of users who would be creating new organizations as well. My thoughts were that 'org admins approve existing org's be  a future iteration, but right now SU is the catchall action.

roverwolf:
Except right now the only button guarded by the super user check is
'Reject', nothing else.

adlearningmate:
@JorgeLu1s 
I added the bootstrap class "btn-danger" and remove the unused CSS .

brandonopened:
Thanks for the feedback Dave, - @adlearningmate please make those changes in permissions:
"both the Suspend and Reject buttons do exactly the same thing, so for a user in the pending state it seems like it would make more sense to show both types of users just Approve and Reject."

PR #404, opened 2018-02-26
Email the admin when there is a new public_user signup

JosephLeon:
When a public_user completes the signup process send an email to the admin notifying them of the user's account.

---
This change is [Reviewable](https://reviewable.io/reviews/opensalt/opensalt/404)


src/App/Command/Email/SendAdminNotificationEmailCommand.php
```diff
@@ -0,0 +1,34 @@
...
+    /**
+     * @var string
+     */
+    protected $userName;
+
+    public function __construct(string $recipient, string $userName, string $userOrganization)
+    {
+        parent::__construct($recipient);
+        $this->userName = $userName;
+        $this->userOrganization = $userOrganization;
```
roverwolf: `$this->userOrganization` should be defined as a class property (like `$this->userName`).

src/App/Handler/Email/AbstractEmailHandler.php
```diff
@@ -36,9 +42,10 @@
      */
     private $mailFromEmail;
 
-    public function __construct(ValidatorInterface $validator, ToggleManager $manager, ContextFactory $contextFactory, \Swift_Mailer $mailer, string $mailFromEmail = null)
+    public function __construct(ValidatorInterface $validator, ToggleManager $manager, ContextFactory $contextFactory, \Swift_Mailer $mailer, string $mailFromEmail = null, TwigEngine $templating)
```
roverwolf: `$templating` should come earlier in the list since `$mailFromEmail` has a default value and might not be passed.

src/Salt/UserBundle/Controller/SignupController.php
```diff
@@ -92,6 +93,18 @@ public function signupAction(Request $request)
                     // Do not throw an error to the client if the email could not be sent
                 }
 
+                // send email to admin about this user creation
+                // get public users username and org
+                try {
+                    $from_email = $this->getParameter('mail_feature_from_email');
+                    $command = new SendAdminNotificationEmailCommand($from_email, $targetUser->getUsername(), $targetUser->getOrg()->getName());
```
roverwolf: Can we assume that the `mail_feature_from_email` is the same email address that should receive notifications about new users?

Can be left for later, but it should probably be documented in an issue so that it is not forgotten about.

src/App/Command/Email/SendAdminNotificationEmailCommand.php
JosephLeon: Added it as a class property.

src/Salt/UserBundle/Controller/SignupController.php
JosephLeon: Added an issue https://github.com/opensalt/opensalt/issues/406

Issue #247, opened 2017-11-10
Human readable export


brandonopened:
We want an export format that is teacher-readable. Greg's wife can provide input on this maybe mine too. CSV/spreadsheet won't cut it. 

Ideally it would be a format more like a PDF where the major statement and description, as well as any notes, could be formatted more like it shows up on the CFItem screen. 
<img width="638" alt="screen shot 2017-11-10 at 11 16 42 am" src="https://user-images.githubusercontent.com/14943151/32674573-ad09fc9e-c608-11e7-8e20-02fabded1895.png">

NileshPatreLM:
We could use a javascript library pdfMake.js to create the pdf with the required data

brandonopened:
Thanks for jumping in Nilesh! http://pdfmake.org/playground.html looks pretty cool. I would say we won't need this feature until sometime in PI6 so good to keep talking about it. Looking at how the pages are rendered now, do you think this might be feasible? It looks like we'd just have to define how the current export data should be formatted, which should be relatively easy since each page is the "same".

jmarks:
Interesting. How would we format the tree into sections and apply pagination and style sheets to different item types? It would be nice to see the markdown and LaTeX in the output and the associations too.

NileshPatreLM:
The steps which we followed to generate the pdf are
1. Insert the case file in database
2. Processing happened at the java side to create the new json response in nested structure(parent-child structure) based on the framework selected 
Ex. Document:{name:"Test"},
     Statements:[{name:"ParentStatement",
                          children:[{name:"Child1"},{name:"Child2"}]                            
}] 

3. The association was represented in nested structure(parent-child structure) in Json file from Java.
4. In the UI , as we already have nesting(still need considerable amount of coding in javascript), we just put it in pdf with left indentation for children (to show parent child relation)

Regarding Latex Expression - there is a narrative field which contains latex expression in HFMS. But I was not able to find this field in the Case file generated for opensalt

If we want to put the latex expression in PDF as well, 
We will need two js libraries
1. Mathjax.js - To parse latex into its mathematical representation
2. html2canvas.js - To print the expression in HTML and capture the image (this image containing math expressions will be inserted in pdf)

brandonopened:
Currently it will export a rough draft. We will have to build in markdown and latex support throughout the next couple weeks. (mathjax works well with the export PDF app so we will use that)

PR #423, opened 2018-03-13
Export Framework as a stylized PDF 

adlearningmate:
Export Framework as a PDF format 
1. Logged in opensalt with user role (Super User, Admin, Editor, Super Editor )
2. Go to the Framework detail page
3. You can see Export button after clicking on it, we will get button for Styled PDF.
which will download the framework details- CFDocuments, CFItems and Association in PDF format

---
This change is [Reviewable](https://reviewable.io/reviews/opensalt/opensalt/423)


roverwolf:
Which GitHub Issue is this for?

roverwolf:
Why are we restricting this from being used by anonymous users?

brandonopened:
It shouldn't be limited to anonymous but available to anonymous... it references #247

roverwolf:
Okay, I also found #137 which sounds like it may apply.

@adlearningmate Based on the comment by @brandonopened it looks like step 1 is incorrect and any user (including anonymous users) should be able to use this feature.

brandonopened:
@adlearningmate yes, let's go ahead and make it available to all users.

PR #427, opened 2018-03-14
Export Framework as a stylized PDF New PR (KMS-3693)

adlearningmate:
Export Framework as a PDF format

Go to the Framework detail page.
You can see Export button after clicking on it, we will get button for Styled PDF.
which will download the framework details- CFDocuments, CFItems and Association in PDF format
Any user (including anonymous users) can use this feature.

Story ID : KMS-3693.  It references git issues [#247](https://github.com/opensalt/opensalt/issues/247) and  [#137 
](https://github.com/opensalt/opensalt/issues/137)
Since there were few issues on old PR, We created this new PR.
We have taken care of all comments from old PR in this new PR.


tests/_support/Page/Framework.php
```diff
@@ -1129,4 +1129,38 @@ public function iUploadTheItemTypeCASEFile()
...
+     * @When /^I download the framework PDF file$/
+     */
+    public function iDownloadTheFrameworkPdfFile(): Framework
+    {
+        $I = $this->I;
+
+        $I->click('Export', '#itemInfo');
+        $I->waitForElementVisible('#exportModal a.btn-export-pdf');
+        $I->click('#pdfgen');
+        $url = $I->grabAttributeFrom('#exportModal a.btn-export-case', 'href');
```
roverwolf: This line is downloading a CASE json file, not the PDF file, so the expectation from the method name is not met (and we never verify that the PDF can be downloaded).

app/Resources/views/framework/ls_doc/_export_component.html.twig
```diff
@@ -2,8 +2,8 @@
...
         <div class="col-xs-3"><a href="{{ path('export_excel_file', {'id': lsDoc.id })}}" role="button" tabindex="3" class="btn btn-primary btn-block btn-export btn-export-excel" style="white-space:normal; height:3.8em"data-toggle="popover" data-trigger="hover" data-placement="bottom" data-html="true" data-content="Exports an Excel spreadsheet file that you can open and edit in spreadsheet programs such as Microsoft Excel.">Spreadsheet export</a></div><!-- {{ path('cfpackage_export2', {'id':lsDoc.id, '_format':'csv'}) }} -->
+        <div class="col-xs-3"><a onclick="SaltLocal.handlePdfFile('{{ lsDoc.uri |local_uri }}.json');" id="pdfgen" role="button" tabindex="3" class="btn btn-primary btn-block btn-export btn-export-pdf" style="white-space:normal; height:3.8em"data-toggle="popover" data-trigger="hover" data-placement="bottom" data-html="true" data-content="Exports a styled PDF that can be printed, easily shared with others, and compared with other versions of the Framework.">Styled PDF</a></div>
```
roverwolf: Since this line was moved and had its tabindex changed to 3, the previous line should have its tabindex changed to 2.

app/Resources/assets/js/lsdoc/index.js
```diff
@@ -437,9 +437,110 @@ var SaltLocal = (function () {
...
+           $.ajax({
+           url: data,
+           type: 'get',
+           dataType: 'json',
+            success: function (response) {
+            var pdfData = response;
+            var pdfMake = require('pdfmake');
+            var pdfFonts = require('pdfmake_fonts');
+            pdfMake.vfs = pdfFonts.pdfMake.vfs;
+    function getCFItemRelationship(identifier){
```
roverwolf: Fixing the indentation within the entire `handlePdfFile` function would be good.  At first the getCFItemRelationship function looks like it is not part of the success function, similar to the rest of the content of the function.

tests/_support/Page/Framework.php
```diff
@@ -1129,4 +1129,18 @@ public function iUploadTheItemTypeCASEFile()
...
+    /**
+     * @When /^I download the framework PDF file$/
+     */
+    public function iDownloadTheFrameworkPdfFile(): Framework
+    {
+        $I = $this->I;
+
+        $I->click('Export', '#itemInfo');
+        $I->waitForElementVisible('#exportModal a.btn-export-pdf');
+        $I->click('#pdfgen');
```
roverwolf: This makes sure that the button is there to click, but there should be a check that a file is downloaded (and that it is a PDF file, possibly also that it contains something expected).

app/Resources/views/framework/ls_doc/_export_component.html.twig
```diff
@@ -2,8 +2,8 @@
...
+        <div class="col-xs-3"><a href="{{ path('export_excel_file', {'id': lsDoc.id })}}" role="button" tabindex="2" class="btn btn-primary btn-block btn-export btn-export-excel" style="white-space:normal; height:3.8em"data-toggle="popover" data-trigger="hover" data-placement="bottom" data-html="true" data-content="Exports an Excel spreadsheet file that you can open and edit in spreadsheet programs such as Microsoft Excel.">Spreadsheet export</a></div><!-- {{ path('cfpackage_export2', {'id':lsDoc.id, '_format':'csv'}) }} -->
+        <div class="col-xs-3"><a onclick="SaltLocal.handlePdfFile('{{ lsDoc.uri |local_uri }}.json');" id="pdfgen" role="button" tabindex="3" class="btn btn-primary btn-block btn-export btn-export-pdf" style="white-space:normal; height:3.8em"data-toggle="popover" data-trigger="hover" data-placement="bottom" data-html="true" data-content="Exports a styled PDF that can be printed, easily shared with others, and compared with other versions of the Framework.">Styled PDF</a></div>
```
roverwolf: An id is probably not required for the button, the clas `btn-export-pdf` is probably enough to distinguish it.

tests/_support/Page/Framework.php
```diff
@@ -1129,4 +1129,18 @@ public function iUploadTheItemTypeCASEFile()
...
+    /**
+     * @When /^I download the framework PDF file$/
+     */
+    public function iDownloadTheFrameworkPdfFile(): Framework
+    {
+        $I = $this->I;
+
+        $I->click('Export', '#itemInfo');
+        $I->waitForElementVisible('#exportModal a.btn-export-pdf');
+        $I->click('//*[@id="exportModal"]/div/div/div[2]/div/div[1]/div/div[3]/a');
```
roverwolf: This makes sure that the button is there to click, but there should be a check that a file is downloaded (and that it is a PDF file, possibly also that the file contains what is expected).

Also, to make it easier to read, you might want to use the same CSS selector that is used in the `waitForElementVisible` call instead of using the XPath selector.

anuradhas248:
@roverwolf - As per your comment, I have replaced x-path to css classes.

roverwolf:
That was only half the comment.

The bigger part of the comment was that the test should check that a file is actually downloaded (and that it is a PDF file, with expected contents).

PR #431, opened 2018-03-16
use an "OpenSALT" version of the node docker image

roverwolf:
Due to the issues we ran into this morning with the bug in the official `node:alpine` docker image, use a forked image that we have more control over (at least temporarily).

---
This change is [Reviewable](https://reviewable.io/reviews/opensalt/opensalt/431)


roverwolf:
It looks like the upstream library is fixed, so closing this for now.

PR #432, opened 2018-03-16
Update spreadsheet export and import

### LINKED ISSUE #409: Add all CASE fields to spreadsheet import  - KMS-3718
Making sure all of the CASE fields are in the spreadsheet importer.
pertains to Editor function - Import spreadsheet framework

EXPLAIN ISSUE OR BUG WITH GHERKIN TEXT

EXAMPLE USER STORY:

AS A user with access to create items  
I WANT the legacy processes and workspace processes to create items in an identical manner  
SO THAT duplicate items and gaps in items do not get generated

JorgeLu1s:
related to #409

---
This change is [Reviewable](https://reviewable.io/reviews/opensalt/opensalt/432)


src/Service/ExcelExport.php
```diff
@@ -167,15 +164,13 @@ public function generateExcelFile(LsDoc $cfDoc, array $items, array $association
...
-            ->setCellValue('F1', 'destinationNodeUri')
-            ->setCellValue('G1', 'associationType')
-            ->setCellValue('H1', 'associationGroupIdentifier')
-            ->setCellValue('I1', 'associationGroupName')
-            ->setCellValue('J1', 'lastChangeDateTime')
+            ->setCellValue('B1', 'originNodeIdentifier')
+            ->setCellValue('C1', 'originNodeHumanCodingScheme')
+            ->setCellValue('D1', 'associationType')
+            ->setCellValue('E1', 'destinationNodeHumanCodingScheme')
+            ->setCellValue('F1', 'destinationNodeIdentifier')
```
roverwolf: Note that without allowing the destination URI an exemplar cannot really be seen nor edited (since the URI is what holds what the exemplar is pointing to)

Issue #434, opened 2018-03-19
There's no underline button in the markdown editor


JosephLeon:
OpenSalt version 1.3.1
All OSs and Browswers

AS A user with edit access to items  
WHEN I edit an item I should see the underline button in the editor

BarbAllen:
KMS-4186

roverwolf:
Resolved by #435

PR #435, opened 2018-03-19
Add underline button and function for the editor.

JosephLeon:
Issue for this PR https://github.com/opensalt/opensalt/issues/434

There is already a test for this functionality however the button was never added to the editor menu.

To verify, go to a framework with items on it, edit an item and check for the underline icon in the editor menu. Highlight some text in the field and click the underline. When clicked it will add underscores around the selected text. The underscores indicate that the item will be underlined when processed as markdown.

---
This change is [Reviewable](https://reviewable.io/reviews/opensalt/opensalt/435)


app/Resources/assets/js/cftree/render-md.js
```diff
@@ -205,18 +205,35 @@ const render = (function(){
...
         return new SimpleMDE({
           element : element,
-          toolbar : [ 'bold', 'italic', 'heading', '|', 'quote',
-            'unordered-list', 'ordered-list', {
+          toolbar : [
+            {
+              name: "underlineText",
+              action: underlineText,
+              className: "fa fa-underline",
+              title: "Underline text (Ctrl/Cmd-Alt-R)",
```
roverwolf: The mentioned shortcut keys do not seem to work.

roverwolf:
I'll also note that if you underline something and then put the cursor into the text that is underlined, the toolbar shows `Italic` highlighted instead of `Underline text`.

JosephLeon:
Added an issue about the visual inconsistency with the visual representation of underlined text vs the output. https://github.com/opensalt/opensalt/issues/498

PR #446, opened 2018-03-23
Pending user should be display on top of the user list page. (KMS-3984)

adlearningmate:
created new user with pending status user should be display on the top of the user list page .
1.login as (SuperUser,Admin)
2.In mange user list page display the all pending status user on top of the list.

---
This change is [Reviewable](https://reviewable.io/reviews/opensalt/opensalt/446)

### DIFF
--- a/src/Controller/User/UserController.php
+++ b/src/Controller/User/UserController.php
@@ -64,8 +64,9 @@ class UserController extends AbstractController
         if ($this->authChecker->isGranted('ROLE_SUPER_USER')) {
             $users = $em->getRepository(User::class)->findAll();
         } else {
+            $user_org = $this->getUser()->getOrg();
             $users = $em->getRepository(User::class)
-                ->findByOrg($this->getUser()->getOrg());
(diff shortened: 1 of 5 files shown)


src/Controller/User/UserController.php
```diff
@@ -64,8 +64,9 @@ public function indexAction()
         if ($this->authChecker->isGranted('ROLE_SUPER_USER')) {
             $users = $em->getRepository(User::class)->findAll();
         } else {
+            $user_org = $this->getUser()->getOrg();
```
JosephLeon: Maybe this should also check of the user has an Org.

tests/acceptance/features/user/super_user/pending_usertopoflist.feature
```diff
@@ -0,0 +1,16 @@
+Feature: Pending new user should be display on top of user list page
+
+    @public-user @user @add-user
```
JosephLeon: Should include a unique identifier like "@7620-0007". You can use a variant of the time and date.

PR #447, opened 2018-03-23
 Export Framework as a stylized PDF from backend

anuradhas248:
It will Download Framework and it's all child items in stylized PDF format. This feature is implemented using PDF Office library. 
Story ID : KMS-3693.  It references git issues [#247](https://github.com/opensalt/opensalt/issues/247) and [#137](https://github.com/opensalt/opensalt/issues/137).

---
This change is [Reviewable](https://reviewable.io/reviews/opensalt/opensalt/447)

PR #455, opened 2018-04-04
Release/2.1

JorgeLu1s:
PRs merged on this branch:

#432 
#449 
#447 
#446 
#443 

Related issues:

#409 
#247 
#137

---
This change is [Reviewable](https://reviewable.io/reviews/opensalt/opensalt/455)


app/AppKernel.php
```diff
@@ -33,7 +33,7 @@ public function registerBundles()
             new Nelmio\CorsBundle\NelmioCorsBundle(),
 
             new Qandidate\Bundle\ToggleBundle\QandidateToggleBundle(),
-            new SimpleThings\EntityAudit\SimpleThingsEntityAuditBundle(),
+            new SimpleThings\EntityAudit\SimpleThingsEntityAuditBundle()
```
roverwolf: The `,` did not have to be removed here.

composer.json
```diff
@@ -91,7 +91,9 @@
 
         "matthimatiker/opcache-bundle": "^0.7",
         "qandidate/toggle-bundle": "^1.1.0",
-        "google/recaptcha": "~1.1"
+        "google/recaptcha": "~1.1",
+        "phpoffice/phpword": "dev-master",
```
roverwolf: We should specify a version instead of `dev-master` so that we won't pull in something unexpected later.

src/Controller/PdfExportController.php
```diff
@@ -0,0 +1,56 @@
...
+    use CommandDispatcherTrait;
+    /**
+     * @Route("/cfdoc/{id}/pdf", name="export_pdf_file")
+     * @Method("GET")
+     *
+     * @param int $id
+     *
+     * @return StreamedResponse
+     */
+    public function exportPdfAction(int $id, Request $request): StreamedResponse
```
roverwolf: The parameter `$request` is not used and could be removed.

tests/_support/Page/User.php
```diff
@@ -335,4 +335,19 @@ public function isRejectTheNewUser()
...
+    /**
+     * @Then /^I see last created user account display top of user list page$/
+     */
+    public function getLastCreatedAccountTopList()
+    {
+        $I = $this->I;
+
+        $username = $I->getRememberedString('lastNewUsername');
+        $I->amOnPage('/admin/user');
+        $I->see($username);
```
roverwolf: The `Then` statement seems to indicate that this should be checking that the user is listed as the top line in the table, not just anywhere.

tests/_support/Page/Framework.php
```diff
@@ -1128,5 +1128,34 @@ public function iUploadTheItemTypeCASEFile()
...
+    /**
+     * @Then /^I should see content in the PDF file$/
+     */
+    public function iShouldSeeContentInThePDFFile(): Framework
+    {
+        $I = $this->I;
+
+        $PdfContent = $I->pdf2text($this->filename);
+        $I->assertNotEmpty($PdfContent, 'PDF file is empty');
+        $I->assertContains($I->getLastFrameworkTitle(), $PdfContent, 'Exported PDF');
```
roverwolf: The displayed error here `Exported PDF` might want to be more explict as to the problem (to make it easier to see when looking at the test output).

src/Entity/User/User.php
```diff
@@ -70,7 +70,8 @@ class User implements AdvancedUserInterface, \Serializable, EquatableInterface
      *      min=8,
      *      max=4096,
      *      minMessage="Password must be at least {{ limit }} charactes long",
-     *      maxMessage="Password cannot be longer than {{ limit }} characters"
+     *      maxMessage="Password cannot be longer than {{ limit }} characters",
```
roverwolf: These rules should probably also go into `AddUserCommand` (currently has a min of 6 chars).

tests/_support/Page/User.php
JorgeLu1s: @adlearningmate I think you worked on this, can you add the missing part to test if the pending user is actually at the top of the table?

tests/_support/Page/Framework.php
JorgeLu1s: @anuradhas248 I think you worked on this, can you please modify the error text?

src/Entity/User/User.php
JorgeLu1s: @adlearningmate I think you worked on this, can you please add the same rules to the `AddUserCommand`?

app/Resources/assets/js/lsdoc/index.js
```diff
@@ -679,7 +679,10 @@ $(document).ready(adjustWindow);
 $(window).on('resize', adjustWindow);
 
 $(document).ready(function () {
-    let table = $('#datatable').DataTable();
+    let table = $('#datatable').DataTable({
+        order: [],
+        columnDefs: [ {orderable: false, targets: [0]}, {orderable: false, targets: [4]} ]
```
roverwolf: This may be why the automated tests are no longer passing, as it seems the column used to sort the table to get the last user added is no longer sortable.

PR #463, opened 2018-04-09
Pdf Export Performance Issue

anuradhas248:
Changed Pdf Export library from TCPDF to MPDF to improve the performance.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -93,7 +93,6 @@
         "qandidate/toggle-bundle": "^1.1.0",
         "google/recaptcha": "~1.1",
         "phpoffice/phpword": "^0.14.0",
-        "tecnickcom/tcpdf": "6.*",
         "mpdf/mpdf": "^7.0"
     },
     "require-dev": {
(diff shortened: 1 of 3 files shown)


roverwolf:
@brandonopened and @jmarks Could you add your opinions as to if including a GPL licenced library like mPDF may be an issue for the OpenSALT MIT licence.  (TCPDF is LGPL)

brandonopened:
Did not catch that.. hmm. 

I feel like may be ok?
https://en.wikipedia.org/wiki/GNU_General_Public_License#Use_of_licensed_software

roverwolf:
The issue might be §2.b) of [GPL2](https://www.gnu.org/licenses/old-licenses/gpl-2.0.html)

> You must cause any work that you distribute or publish, that in whole or **in part contains** or is derived from the Program or any part thereof, to be licensed as a whole at no charge to all third parties **under the terms of this License [GPL2]**.

roverwolf:
It looks like we cannot use mPDF due to it having a [GPL](https://github.com/mpdf/mpdf/blob/development/LICENSE.txt) licence.  mPDF is not a build-only tool nor is it "arms length" from the rest of the running application thus it cannot be used as part of OpenSALT without OpenSALT also being licenced as GPL.

From [GPL2 FAQ](https://www.gnu.org/licenses/old-licenses/gpl-2.0-faq.html#MereAggregation):
> If modules are designed to run linked together in a shared address space, that almost surely means combining them into one program.

PR #473, opened 2018-04-18
Spreadsheet updater

### LINKED ISSUE #410: Spreadsheet Import with Update  KMS-3775
As an OS Editor  I WANT to update a framework from a spreadsheet so that Identifiers will be overwritten, Identifiers not found in the import will be deleted and import identifiers not present in the import will be added.

EXPLAIN ISSUE OR BUG WITH GHERKIN TEXT

EXAMPLE USER STORY:

AS A user with access to create items  
I WANT the legacy processes and workspace processes to create items in an identical manner  
SO THAT duplicate items and gaps in items do not get generated

JorgeLu1s:
closes #410

---
This change is [Reviewable](https://reviewable.io/reviews/opensalt/opensalt/473)


src/Service/ExcelImport.php
```diff
@@ -271,4 +312,27 @@ private function getCellValueOrNull(Worksheet $sheet, int $col, int $row)
 
         return $cell->getValue();
     }
+
+    private function checkRemovedElements($doc, $array, $type)
```
roverwolf: A better name than `$array` would be good here (is this something like $keepIdentifierList ?).

JorgeLu1s: this is an array of items or associations from the spreadsheet file, every item/association not present in this array will be deleted from the database. Due `$existingItems` refer to elements from the database do you think `$keepIdentifierList` would be a good name? or something like `$keepElementsList`?

roverwolf: Either name sounds fine as it lets one know a bit about what it is for.  `$array` tells us almost nothing.

Issue #393, opened 2018-02-20
Adding support for attaching images, files and videos


brandonopened:
Currently, it is not possible to attach a file, image or video to OpenSALT fullstatement, notes or exemplar.

- [ ] OS Adding ICON for attachment for full statement, and notes   #262  PI7
    As a user, I want to be able to add an attachment to my comments when clicking an icon in the toolbar.
    
    Description: Icons for file & video, and an icon for pictures is desirable.
    
     
    
     
    
        
     
    
- [ ] OS Adding ICON for attachment for  Exemplar  #262  PI7
    As a user, I want to be able to add an attachment to description in Exemplar.
    
    Description: For Exemplar - a button to attach a file, picture or a video
    
     
    
     
    
     
    
     
    
        
     
    
- [ ] OS Enable File Storage  (attachment upload) PI7 #191
     Description:One possible solution is to use github file storage when users upload an image - This is related to both [#106](https://github.com/opensalt/opensalt/issues/106) and [#147](https://github.com/opensalt/opensalt/issues/147) (previous thought was for an AWS bucket but this seems better).
    
     making file storage configurable, how are we going to do that?
    
    concern - moving from internal and external - file transfer needs to be worked out.
    
    button on attachment for add and remove of attachment
    
     
    
     
    
      
    
    | As |  I WANT TO  | SO THAT |
    | --- | --- | --- |
    
- [ ] OS Adding attachment for full statement, notes - FILE or VIDEOS #262  PI7
     .Assumed that icons will be on the edit bar for Fullstatement and Notes, and available for Exemplar. There is a message for unrecognized file formats, if it is feasible to put the message in a tool tip, that would be good too. In that case the message would be same.
    
     
    
      
    
    | As an OS Editor |  I WANT TO edit Full Statement, Notes | SO THAT I can  attach a file or a video. |
    
- [ ] OS Adding attachment NOT NEEDED - PICTURES #262  PI7
    Markdown supports in-line pictures. 
    
    Description: As a user, I should be able to add an attachment to my comments when editing full statement, notes or exemplar.  The best outcome is that pictures can be attached or displayed on-line.
    
      
    
    | As an OS Editor |  I WANT TO edit Full Statement, Notes or Exemplar description | SO THAT I can place a PICTURE |
    
- [ ] OS Adding attachment for full statement, notes and Exemplar  VALIDATIONS #262  PI7
[...]

PR #481, opened 2018-04-24
Attachment Storage on AWS

anuradhas248:
Created temporary UI for upload files(images, videos, files) to S3 bucket. Tested aws connectivity on dev server. It works on ACT internal instance.
Story ID= KMS-3947
Git Issue = [393](https://github.com/opensalt/opensalt/issues/393)

---
This change is [Reviewable](https://reviewable.io/reviews/opensalt/opensalt/481)


src/Controller/AwsStorage.php
```diff
@@ -0,0 +1,95 @@
...
+use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;
+use Symfony\Component\HttpFoundation\StreamedResponse;
+use League\Flysystem\AwsS3v3\AwsS3Adapter;
+use League\Flysystem\Filesystem;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\Form\Extension\Core\Type\FileType;
+use Symfony\Component\Form\Extension\Core\Type\SubmitType;
+use Aws\Credentials\CredentialProvider;
+
+class AwsStorage extends AbstractController
```
JorgeLu1s: This class must be renamed and end with the word 'Controller'

src/Controller/AwsStorage.php
```diff
@@ -0,0 +1,95 @@
...
+            }
+        }
+       $aws_data = $filesystem->listContents();
+        return $this->render('framework/doc_tree/aws_upload.html.twig', array(
+            'form' => $form->createView(),
+            'aws_files' => $aws_data,
+        ));
+    }
+
+    private function configuration()
```
JorgeLu1s: Maybe this can be moved to a service class in the `src/Service` dir. @roverwolf what do you think?

src/Entity/Framework/AwsStorage.php
```diff
@@ -0,0 +1,145 @@
...
+     * @ORM\Column(name="file_name", type="string", length=250)
+     *
+     * @Assert\NotBlank
+     */
+    protected $fileName;
+
+    /**
+     * Constructor
+     */
+    public function __construct()
```
JorgeLu1s: I think this construct can be removed since it is empty

src/Controller/AwsStorage.php
```diff
@@ -0,0 +1,95 @@
...
+    {
+        $provider = CredentialProvider::defaultProvider();
+        $client = \Aws\S3\S3Client::factory([
+        'credentials' => $provider,
```
roverwolf: The `defaultProvider` is the default, so this line (and lien 56) are not needed if no other credential options will be supported.

src/Controller/AwsStorage.php
roverwolf: Yes, the configuration should be a service, possibly just using passed environment variables and defining services in the services.yml file, or if more complicated then using services to do the configuration.

Since this is also the *only* place that specifies that this is using S3 we might want to rename the classes to change the Aws portion and allow for the possibility of using other adapters as well, with S3 being the only supported adapter (or maybe allow the local adapter to make it easier for developers?  or developers could use Minio for testing with?).

JorgeLu1s:
Closing this PR because of Vish request

PR #482, opened 2018-04-25
Change password criteria is not working properly (KMS-4225)

adlearningmate:
Reset password criteria is not working properly when trying to reset the password from User Management screen.
1- Login by the Super User.
2- Manage user - select edit button from any user from the list.
3- User edit window - try to change the password without password criteria.

Login in opensalt with any user .
1. click the right side tab(ex:abc@gmail.com)
2. than click the change password link.
3. insert the old password , New Password and Repeat Password.

---
This change is [Reviewable](https://reviewable.io/reviews/opensalt/opensalt/482)

### DIFF
--- a/src/Controller/User/ChangePasswordController.php
+++ b/src/Controller/User/ChangePasswordController.php
@@ -43,7 +43,7 @@ class ChangePasswordController extends AbstractController
     public function changePasswordAction(Request $request)
     {
         $dto = new ChangePasswordDTO();
-        $form = $this->createForm(ChangePasswordType::class, $dto);
+        $form = $this->createForm(ChangePasswordType::class, $dto, ['validation_groups' => ['registration']]);
 
         $form->handleRequest($request);
 
--- a/src/Controller/User/ChangePasswordController.php
+++ b/src/Controller/User/ChangePasswordController.php
@@ -43,7 +43,7 @@ class ChangePasswordController extends AbstractController
     public function changePasswordAction(Request $request)
     {
         $dto = new ChangePasswordDTO();
-        $form = $this->createForm(ChangePasswordType::class, $dto, ['validation_groups' => ['registration']]);
+        $form = $this->createForm(ChangePasswordType::class, $dto, ['validation_groups' => ['changepass']]);
 
         $form->handleRequest($request);
 
--- a/src/Controller/User/UserController.php
+++ b/src/Controller/User/UserController.php
@@ -154,7 +154,7 @@ class UserController extends AbstractController
     public function editAction(Request $request, User $targetUser)
     {
         $deleteForm = $this->createDeleteForm($targetUser);
-        $editForm = $this->createForm(UserType::class, $targetUser, ['validation_groups' => ['registration']]);
+        $editForm = $this->createForm(UserType::class, $targetUser);
         $editForm->handleRequest($request);
 
         if ($editForm->isSubmitted() && $editForm->isValid()) {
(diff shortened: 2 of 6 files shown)


src/Controller/User/ChangePasswordController.php
```diff
@@ -43,7 +43,7 @@ public function __construct(UserPasswordEncoderInterface $passwordEncoder)
     public function changePasswordAction(Request $request)
     {
         $dto = new ChangePasswordDTO();
-        $form = $this->createForm(ChangePasswordType::class, $dto);
+        $form = $this->createForm(ChangePasswordType::class, $dto, ['validation_groups' => ['registration']]);
```
roverwolf: Why are we using a `registration` group here?  I'm not sure why we would need any validation group.
- Note that using a group named`registration` does not make sense here as the form is for changing a password and has nothing to do with registering a user.

Leaving it off also means no changes would need to be made to `Form/Type/ChangePasswordType.php` instead of adding a validation group there as well, and `Form/DTO/ChangePasswordDTO.php ` would also not have a validation group specifiied in the three locations it is now (by this PR) being specified there.

src/Controller/User/UserController.php
```diff
@@ -154,7 +154,7 @@ public function showAction(User $targetUser)
     public function editAction(Request $request, User $targetUser)
     {
         $deleteForm = $this->createDeleteForm($targetUser);
-        $editForm = $this->createForm(UserType::class, $targetUser);
+        $editForm = $this->createForm(UserType::class, $targetUser, ['validation_groups' => ['registration']]);
```
roverwolf: This should **not** have the validation group `registration` added as adding it means that a new password for the user would be **required** when any *other change* to the user is made (such as changing their role).  One should be able to change the user's role without changing their password.

src/Form/Type/ChangePasswordType.php
```diff
@@ -18,7 +18,7 @@ public function buildForm(FormBuilderInterface $builder, array $options)
             ->add('newPassword', RepeatedType::class, [
                 'type' => PasswordType::class,
                 'invalid_message' => 'The password fields must match',
-                'required' => true,
+                'required' => in_array('changepass', $options['validation_groups'], true),
```
roverwolf: This form is specifically for changing passwords, so it does not make sense for the newPassword to ever not be required.

src/Form/DTO/ChangePasswordDTO.php
```diff
@@ -4,23 +4,29 @@
...
 use Symfony\Component\Validator\Constraints as Assert;
+use App\Validator\Constraints as CustomAssert;
 
 class ChangePasswordDTO
 {
     /**
      * @var string
      *
-     * @SecurityAssert\UserPassword(message="The current password is incorrect")
+     * @SecurityAssert\UserPassword(message="The current password is incorrect", groups={"changepass"})
```
roverwolf: The ChangePasswordDTO class is only for changing passwords, so adding a group for the validation does not make much sense as there are no variations in how the validation is checked.

src/Controller/User/ChangePasswordController.php
```diff
@@ -43,7 +43,7 @@ public function __construct(UserPasswordEncoderInterface $passwordEncoder)
     public function changePasswordAction(Request $request)
     {
         $dto = new ChangePasswordDTO();
-        $form = $this->createForm(ChangePasswordType::class, $dto);
+        $form = $this->createForm(ChangePasswordType::class, $dto, ['validation_groups' => ['changepass']]);
```
roverwolf: There is no need to have validation groups for the change password form.

Issue #505, opened 2018-05-18
Alt text is appearing along with inline images on exported pdf file


sajeed42:
Steps:

    Select Framework with an Image
    Select Export button
    Select Styled PDF

Actual:

Alt text is appearing along with inline images on exported pdf file

Expected:

Inline images should appearing without showing alt text on exported pdf file
![image alt text](https://user-images.githubusercontent.com/33145683/40231006-1e4835f4-5ab7-11e8-904c-1cf894b04fbd.png)

PR #506, opened 2018-05-18
Fixed 'alt text was rendering with inline image in pdf' issue

anuradhas248:
Fixed 'alt text was rendering with inline image in pdf' issue and also allowed http with https for image url.

Jira ID - KMS-4389
Git Issue - [505](https://github.com/opensalt/opensalt/issues/505)

---
This change is [Reviewable](https://reviewable.io/reviews/opensalt/opensalt/506)

### DIFF
--- a/src/Controller/PdfExportController.php
+++ b/src/Controller/PdfExportController.php
@@ -69,8 +69,8 @@ class PdfExportController extends Controller
      */
     public function render_images($string)
     {
-        $pattern = '/\!\[([^\]]*)\]\(((?:https:\/\/|\/)[^\)]+)\)/';
-        $result = preg_replace($pattern, '$1 <img src = "$2">', $string);
+        $pattern = '/\!\[([^\]]*)\]\(((?:(?:https?|):\/\/|\/)[^\)]+)\)/';
+        $result = preg_replace($pattern, '<img src = "$2">', $string);
         return $result;
     }
 }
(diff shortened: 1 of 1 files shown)


src/Controller/PdfExportController.php
```diff
@@ -69,8 +69,8 @@ function () use ($phpWordObject) {
      */
     public function render_images($string)
     {
-        $pattern = '/\!\[([^\]]*)\]\(((?:https:\/\/|\/)[^\)]+)\)/';
-        $result = preg_replace($pattern, '$1 <img src = "$2">', $string);
+        $pattern = '/\!\[([^\]]*)\]\(((?:(?:https?|):\/\/|\/)[^\)]+)\)/';
```
roverwolf: What was the intention of this change?  Do we really want to allow `://something-here` to be valid?  Or was the intention to just add the `?` after the `s` to allow `http`?

roverwolf: Or was the pattern supposed to change at all?  Just the replacement (removing $1)?

anuradhas248: Testers have found few urls that used pattern alt text. And http was not allowing in previous pattern. So for allow http and pattern similar like "://something-here." I updated pattern. At the same time I removed $1 as it was showing first bracket text along with image.

anuradhas248:
@roverwolf - As per your suggestions, I have updated pattern that allowing http(as some of content used http url to render image) along with https. And removed second condition which was allowing :// prefix.
Please verify.
Thank you.

Issue #545, opened 2018-06-21
Add TaskModel


JosephLeon:
TaskModels allow for additional information to be recorded for an Item. Data scientists use TaskModels.

### Feature story
```
Feature: TaskModel
  As a user
  I want be able to CRUD task models
  So that task models can be used in OS
```

PR #546, opened 2018-06-21
Add TaskModel Entity

JosephLeon:
https://github.com/opensalt/opensalt/issues/545

Data Scientists need a way to manage Task Models in OS. This PR allows users to CRUD Task Models and relate them to an LsItem.

---
This change is [Reviewable](https://reviewable.io/reviews/opensalt/opensalt/546)


brandonopened:
@JosephLeon can we close this or will we need this code in our new approach to this feature?

Issue #541, opened 2018-06-13
PI8 Adding support for attaching images, files, and videos


brandonopened:
- [ ] OS Ability to add Inline Image via UI toolbar button #262 PI8
    Need to complete SPIKE KMS-4283: researching inline images before pointing
    
     
    
    As an OS editor  
     I want to be able to select a button to insert an inline image    
     So that I can add images to the Full Statement and Notes fields in 'Add' or 'Edit' modes
    
    **SCOPE**  
    The scope of this story is to allow users to click on a button in the toolbar in order to add embedded images to Full Statement or Notes fields.  Per Barb's note - Looks like code mirror doesn't allow alt text
    
    **OUT OF SCOPE**  
     Adding inline images to the Comments section is out of scope at this time.  Attachments story may take care of this (KMS-3929)
    
    Adding inline images via 'Import Children' is out of scope at this time
    
    **AUTOMATED ACCEPTANCE TEST CASES**  
     < need input from team>
    
    **DOMAIN**  
     Login to OS Staging environment:  [ [http://frameworks-staging.act.org/login](http://frameworks-staging.act.org/login)]
    
    - OS > Select Framework > Select Parent or Child Item > 'Edit' button > Full Statement AND Notes text box
    - OS > Select Framework > Select 'Add New Child Item' button > Full Statement AND Notes text box
    - OS > Select Framework > Select 'Export' button 
    - OS > Select Framework > Select 'Copy Framework' button  
      
    
    **DOCUMENTATION**
    
    [https://opensalt.mybalsamiq.com/projects/opensalt/CF%20Item%20Edit](https://opensalt.mybalsamiq.com/projects/opensalt/CF%20Item%20Edit)
    
- [ ] OS Attachment in Commenting:  Enable S3 bucket for external site #262 PI8
    Need to complete SPIKE KMS-4293 before this story
    
     
    
    As an OS user on the external site  
     I want the system updated  
     So that I can attach files to my comments
    
    **SCOPE**  
     The scope of this story is to enable the S3 bucket for the external site ( [http://frameworks-staging.act.org/](http://frameworks-staging.act.org/) and frameworks.act.org) in order to allow these users to attach files to comments.
    
    **OUT OF SCOPE**  
     KMS-3929 will update the system configuration to allow commenting with attachments
    
    **AUTOMATED ACCEPTANCE TEST CASES**  
    Repeat the test cases that the file is downloaded from S3 which by default also tests that it was uploaded
    
[...]

Issue #525, opened 2018-05-29
Dev Enable S3 bucket for external site for Commenting


Vishvesh-LM:
Research getting ACT's S3 bucket to work with external servers (Staging and Production)

roverwolf:
Resolved by #549

Issue #494, opened 2018-05-08
Attachment in Commenting 


Vishvesh-LM:
**Description**
As an OS reviewer
I would like commenting updated
So that I can click to attach a file to my comments

**SCOPE**

The scope of this story is to update commenting to allow OS reviewers the ability to attach a file to their comments via an 'Attachment' button.  This needs to be turned on via a configuration. The scope also includes the ability to remove the attachment.

- Supported file formats:  gif, jpg, jpeg, png, tiff, tif, svg
- No limit on file size (unless technical constraints are identified)
- User should be able to export comments with a URL to the image -- the image should be available externally
- Maximum of 3 attachments to a comment and 3 attachments to a reply 
- Attachments will display with the comments/reply box
- Any user with the ability to comment/reply should be able to add an attachment
- Current AC indicate attachment will display in a new tab (prior stories with pdf, doc./docx indicated download them other file types allow direct access)
- After uploading, the user would see the attachment, if successful, or receive a pop up message if it failed.  This is part of the commenting functionality and will just need to be sure it still occurs.

OUT OF SCOPE
No items are defined as out of scope at this time

AUTOMATED ACCEPTANCE TEST CASES
Confirm the files exist and up to 3 are allowed in both comments or a reply
Confirm supported file formats and ensure unsupported are not allowed

DOMAIN
Select a framework > Add attachment in comments box

ACCEPTANCE CRITERIA

Please see file attached

[KMS3929_AcceptanceCriteria.docx](https://github.com/opensalt/opensalt/files/1984836/KMS3929_AcceptanceCriteria.docx)

Issue #490, opened 2018-05-03
User try to attached files are not getting uploaded and saved on fullstatement and notes fields


sajeed42:
Steps:

    login with superuser on https://fds-dev.corporate.act.org/
    click on any framework
    click add new child item button
    try to upload by click on upload attachment field on both full statement and notes fields
    verify some files get upload and most does not get upload
    click on save for new create and for editing existing statement

Actual:

Files are not getting uploaded and saved

Expected:

Files should get uploaded and saved

PR #549, opened 2018-07-03
Added Flysystem library to handle Aws s3 Bucket.

JorgeLu1s:
Created a service to upload attachments to s3.
Added code to upload attachments from the commenting system.

---
This change is [Reviewable](https://reviewable.io/reviews/opensalt/opensalt/549)


roverwolf:
Relates to #476 #490 #494 #525 #541

src/Controller/CommentsController.php
```diff
@@ -254,16 +255,25 @@ private function url(string $itemType, Comment $comment): ?string
      *
      * @return JsonResponse
      */
-    private function addComment(Request $request, string $itemType, $item, UserInterface $user): Response
+    private function addComment(Request $request, string $itemType, $item, UserInterface $user, Aws $aws): Response
```
roverwolf: `$aws` can probably be added to the constructor and not need to be passed here (or into `newDocCommentAction`.

src/Entity/Comment/Comment.php
```diff
@@ -109,6 +109,20 @@ class Comment
      */
     private $updatedAt;
 
+    /**
+     * @var string
+     *
+     * @ORM\Column(type="text")
+     */
+    private $fileUrl;
```
roverwolf: Should this (and `$fileMimeType`) be nullable?

src/Service/Aws.php
```diff
@@ -0,0 +1,64 @@
...
+
+    public function __construct(string $awsKey = null, string $secret = null, string $bucket = null, string $prefix = null)
+    {
+        $this->awsKey = $awsKey;
+        $this->secret = $secret;
+        $this->bucket = $bucket;
+        $this->awsPrefix = $prefix;
+    }
+
+    public function filesystem () {
```
roverwolf: This file should probably be split  into two, one for managing the files and the second part to configure the adapter (and injected into the first).

assets/js/lsdoc/comments.js
```diff
@@ -99,22 +100,63 @@ var CommentSystem = (function(){
...
+
+                    $(Object.keys(commentJSON)).each(function (index, key) {
+                        var value = commentJSON[key];
+                        if (value) {
+                            formData.append(key, value);
+                        }
+                    });
+
+                    $.ajax({
+                        url: '/app_dev.php/comments/'+commentItem.itemType+'/'+commentItem.itemId,
```
roverwolf: `app_dev.php` should not be hardcoded here.

docker/docker-compose.test.yml
```diff
@@ -45,7 +45,13 @@ services:
...
       GOOGLE_CAPTCHA_SITE_KEY:
       GOOGLE_CAPTCHA_SECRET_KEY:
-    volumes_from:
+      AWS_ACCESS_KEY_ID:
+      AWS_SECRET_ACCESS_KEY:
+      AWS_S3_PREFIX:
+      AWS_S3_BUCKET:
+      AWS_CLOUDFRONT_DOMAIN:
+      BUCKET_PROVIDER:
+      volumes_from:
```
roverwolf: `volumes_from:` seems to have gotten an extra indent here which needs to be removed.

src/Service/BucketFactory.php
```diff
@@ -0,0 +1,39 @@
...
+
+    private $params;
+
+    public function __construct(ParameterBagInterface $params)
+    {
+        $this->params = $params;
+    }
+
+    public function filesystem () {
+        $adapter = new Local('/tmp/attachments');
```
roverwolf: The local filesystem location should be configurable.

src/Service/BucketFactory.php
```diff
@@ -0,0 +1,39 @@
...
+
+        if ($this->params->get('bucket_provider') == 'S3') {
+            $client = new S3Client([
+                'credentials' => [
+                    'key' => $this->params->get('aws_key'),
+                    'secret' => $this->params->get('aws_secret'),
+                ],
+                'region' => 'us-east-1',
```
roverwolf: The AWS region should be configurable.

src/Service/BucketService.php
```diff
@@ -0,0 +1,39 @@
...
+{
+    private $filesystem;
+
+    public function __construct(Filesystem $filesystem) {
+        $this->filesystem = $filesystem;
+    }
+
+    public function uploadFile ($file, $dir) {
+        $filesystem = $this->filesystem;
+        $path = "/$dir/".$file->getClientOriginalName();
```
roverwolf: The path should probably be generated to be something other than the original file name as otherwise 1) different comments can't have the same filename uploaded, 2) the extension might not match the filename correctly (possible security issue)

src/Service/BucketFactory.php
```diff
@@ -0,0 +1,39 @@
...
+    public function __construct(ParameterBagInterface $params)
+    {
+        $this->params = $params;
+    }
+
+    public function filesystem () {
+        $adapter = new Local('/tmp/attachments');
+
+        if ($this->params->get('bucket_provider') == 'S3') {
+            $client = new S3Client([
```
roverwolf: Probably should allow for other options to be passed as well, such as `endpoint` and `use_path_style_endpoint` (to be able to test with S3-compatible APIs such as Minio).

src/Service/BucketFactory.php
```diff
@@ -0,0 +1,39 @@
...
+                'credentials' => [
+                    'key' => $this->params->get('aws_key'),
+                    'secret' => $this->params->get('aws_secret'),
+                ],
+                'region' => 'us-east-1',
+                'version' => 'latest'
+            ]);
+
+            $adapter = new AwsS3Adapter($client, $this->params->get('aws_bucket'), $this->params->get('aws_prefix'));
```
roverwolf: Would be better to just `return new AwsS3Adapter()` here and then `return new Local()` as a default at the end instead of setting that first.

assets/js/lsdoc/comments.js
```diff
@@ -1,24 +1,25 @@
...
     }
 
-    function init(nodeRef){
+    function init (nodeRef){
         setItem(nodeRef);
         $('.js-comments-container').comments({
             profilePictureUrl: '',
             enableDeletingCommentWithReplies: true,
-            getComments: function(success, error) {
+            enableAttachments: true,
```
roverwolf: I thought attachments is supposed to be an option, so a way to disable it is necessary.

src/Service/BucketService.php
```diff
@@ -0,0 +1,39 @@
...
+        fclose($stream);
+
+        return $path;
+    }
+
+    public function retrieveFile ($path) {
+        $filesystem = $this->filesystem;
+
+        $stream = $filesystem->readStream($path);
+        $contents = stream_get_contents($stream);
```
roverwolf: Are the `$contents` supposed to be used somewhere?

JorgeLu1s: Actually this function is not needed since we can access to the file directly with the cloudfront domain

assets/js/lsdoc/comments.js
JorgeLu1s: you're right

src/Controller/CommentsController.php
```diff
@@ -254,16 +255,25 @@ private function url(string $itemType, Comment $comment): ?string
...
 
         $parentId = $request->request->get('parent');
         $content = $request->request->get('content');
+        $fileUrl = null;
+        $fileMimeType = null;
 
-        $command = new AddCommentCommand($itemType, $item, $user, $content, (int) $parentId);
+        $file = $request->files->get('file');
+
+        if ($file->isValid()) {
```
roverwolf: This should probably be checking that the comment_attachments feature is enabled before trying to upload a file.

roverwolf:
PR #566 needs to be applied so that `composer` will work in travis and scrutinizer (symfony/flex was updated to work with the latest version of composer).

JorgeLu1s:
@roverwolf I've added the content validation for the file. If there is needed to set up some environment variables in order to make the test pass, how can I do that? I mean here in the CI

roverwolf:
@JorgeLu1s You would have to set the environment variables to enable comments and comment attachments, the same as you have done on your local system to run the tests there.

Issue #563, opened 2018-07-26
Change the way the URI is displayed and linked in the item detail panel


jmarks:
The item detail has the "CASE Framework URI" with a link. However this is the OpenSALT URI not the CASE URI, but both are useful. Instead display just the UUID as the "Identifier" and have it link to the CASE URI so you can copy the CASE URI.

roverwolf:
@jmarks Note that the "CASE URI" is not required by the spec to be a URL to the CASE API endpoint.

brandonopened:
So, with #578 our item detail changed a bit! 
We just need to add CASEItemGUID underneath it. Will help with implementation/demo for LTI Resource Search etc.

brandonopened:
Ok, it looks like we would add it here:
  for (key in attributes = {
                'uri': 'CASE Item URI',
                'le': 'List Enumeration in Source',
                'cku': 'Concept Keywords URI',
                'lang': 'Language',
                'licenceUri': 'Licence URI',
                'mod': 'Last Changed'

on opensalt/assets/js/cftree/view-documentclass.js ? 

I tried locally adding in identifier, UUID, and GUID but they didn't work. Probably is something that we would need to change in two places plus add a test.

brandonopened:
What I want to see is the identifier alone like on the LSDoc index page. that uses this variable: lsDoc.identifier but i don't see where to put it on the CfTree item view.

jmarks:
Update on design for this:
Display should include two lines:
Identifier:  UUID as a number that can be copied
URI: the actual URI with a hyperlink. 

[Perhaps the UUID could be a hyperlink to the URI to do this in a single line]

brandonopened:
Dev branch looks like this:
<img width="587" alt="screen shot 2018-11-27 at 2 37 59 pm" src="https://user-images.githubusercontent.com/14943151/49116408-498cbb80-f252-11e8-9024-5921d330a468.png">
<img width="638" alt="screen shot 2018-11-27 at 2 37 55 pm" src="https://user-images.githubusercontent.com/14943151/49116409-498cbb80-f252-11e8-9380-a2ee058414c2.png">



However, when I load a CASE framework from an external source, it correctly has a local URI throughout the framework that works with the API. 
<img width="826" alt="screen shot 2018-11-27 at 2 39 27 pm" src="https://user-images.githubusercontent.com/14943151/49116415-4db8d900-f252-11e8-8369-1f25108ec37b.png">

[...]

PR #578, opened 2018-08-14
Use server-provided value for "Case Framework URL"

### LINKED ISSUE #471: URI links not working always?
on http://frameworks.act.org/cftree/item/19974
http://frameworks.act.org/uri/d833a66c-2176-4240-b4fa-320dee00e87b
the CASE item URI link doesn't work.

I cannot replicate this bug with other frameworks; is it because we loaded this framework over a year ago? 

(URI links on opensalt.net etc seem to work)

### LINKED ISSUE #568: CASE Format URL does not work on imported Frameworks
### Describe the problem
The "CASE Format URL" link doesn't work in Frameworks that were created by importing a CASE file.

### Version of OpenSALT
2.1.0

### To reproduce
1. Select Import Framework
2. Select and import either of the CASE files in the attached ZIP
3. Click on the newly imported Framework
4. In the Item Details pane, click the CASE Format URL link

[case-format-examples.zip](https://github.com/opensalt/opensalt/files/2271279/case-format-examples.zip)

### Expected behaviour
The link should redirect you to the imported framework in your OpenSALT instance.

### Actual behaviour
You receive an error "Could not find the Uri: <identifier>"

### Your system (please complete the following information):
 - Device: desktop
 - OS: multiple
 - Browser: multiple

### Additional context

Related: #471, #563

Database results from my import (showing how this issue happens even when there are no URI's in the CASE file):

```
select  id, uri, identifier, title 
from    ls_doc 
where   identifier in ('6635643b-6c2f-4a28-9713-f52190b28ff9', '4be8c04b-ce77-4b6a-8084-f28a3c1d010a');
```

| id | uri                                                                     | identifier                           | title                                            |
|----|-------------------------------------------------------------------------|--------------------------------------|--------------------------------------------------|
|  7 | https://www.testframeworks.org/doc/6635643b-6c2f-4a28-9713-f52190b28ff9 | 6635643b-6c2f-4a28-9713-f52190b28ff9 | Test Framework with Imported URIs                |
| 11 | local:285619f8-9b26-11e8-96b7-0242ac1a0003                              | 4be8c04b-ce77-4b6a-8084-f28a3c1d010a | Test Framework Without Imported URIs             |

jonscheiding:
Closes #568
Closes #471

Related #563

Updates client-side code to display the server-provided `uri` property value for "Case Framework URL".

**Test plan**

Manual testing:

1. Import the attached CASE file
2. Browse to the "Test Framework with Imported URIs" Framework
3. Expect that the Document's "Case Framework URL" is **http://localhost:3000/uri/https://www.testframeworks.org/doc/1**
4. Expect that the Item's "CASE Item URI" (under "More Info") is **http://localhost:3000/uri/https://www.testframeworks.org/item/1**

[case-format-with-uri.json.zip](https://github.com/opensalt/opensalt/files/2287979/case-format-with-uri.json.zip)

screen shot 2018-08-14 at 5 08 13 pm

---
This change is [Reviewable](https://reviewable.io/reviews/opensalt/opensalt/578)


jmarks:
@jonscheiding Can you add a screenshot of the link look, I had some ideas on this associated with #563.

I was thinking the UUID could drop down a link of related API and URI links so you can copy them.

jonscheiding:
@jmarks I added a screenshot.  I didn't do anything so interesting as what you suggest ... just updated to meet what seemed to be the intent of the current implementation.

If I understand you correctly, you're looking to do something along these lines?

screen shot 2018-08-14 at 5 23 46 pm

(Excuse the crappy mockup.)

jmarks:
@jonscheiding Yes that is the idea, so you can copy the links. This looks to be for doc, so we would need similar for item. And the CASE API shouls use the full IMS URI pattern (Maybe for 1.0 and 1.1 when there is a 1.1) and offer the OpenSALT URI as an option. 

Also please note for the doc, the ability to set a "URL name" which would affect the URI.

jonscheiding:
@jmarks thanks for the feedback.  Your ideas are a bit outside what I'm comfortable trying to implement with my current level of domain understanding.  Would it be possible to move forward with this fix, which is much narrower in scope?  I (hopefully) am not doing anything interesting here ... just using the URI that the server constructed, rather than constructing it separately (with different logic) on the client.

The Travis build seems to have stalled, rather than failed ... is there a way I can re-trigger it?

roverwolf:
@jonscheiding The build did stall, but only after it had already failed.  The reason for the failure is due to travis not being able to access the encrypted environment variables when trying to deal with the docker images.  I'll look at fixing the problem (#585).

jmarks:
@jonscheiding David will work with you on a scope for this, btu lets leave the full scope open.

jonscheiding:
@roverwolf Thanks for the feedback!

I made a few fixes in response to your comment.  Let me know what you think of them.

1. Return `uri` in the `CFItem` objects
2. Make client-side fall back to the previous logic if it doesn't get a URI
3. Make `/uri/<identifier>` work for non-local items (as you suggested in the related issue)

#3 in particular is kind of beyond the scope of what I was originally trying to accomplish so I will happily revert it if you like.

jmarks:
@jonscheiding Thank you for the commits. It is fine to just fix the broken external URI issue and not change the presentation at this time. But let's keep the presentation enhancement you mocked-up here in the backlog and maybe you can take that one at another time.

roverwolf:
URIs for items from different sources appear to be working.  I've merged the code.

PR #579, opened 2018-08-14
Set identifiers in constructor when importing CASE files

jonscheiding:
#568

Updates the import handler to pass identifiers (if they exist) into the constructors for `LsDoc`/`LsItem`/`LsAssociation`, so that they will be used in the automatically-generated `local:` URI.

**Test plan (required)**

Manual testing:
1. Import the attached CASE file.
2. Browse to the "Test Framework Without Imported URIs" Framework
3. Expect that the Document's "Case Framework URL" is **http://localhost:3000/uri/4be8c04b-ce77-4b6a-8084-f28a3c1d010a**
4. Expect that the Item's "CASE Item URI" (under "More Info") is **http://localhost:3000/uri/194a6995-712a-4972-826c-b3d79e92c2ff**

---
This change is [Reviewable](https://reviewable.io/reviews/opensalt/opensalt/579)


roverwolf:
@jonscheiding Thank you for your contribution.

PR #598, opened 2018-08-31
feat: add left footer text

roverwolf:
Allow text to be added to the left side of the footer.
The content of the environment variable FOOTER_LEFT_TEXT will be
displayed.

---
This change is [Reviewable](https://reviewable.io/reviews/opensalt/opensalt/598)

### DIFF
--- a/assets/sass/styles.scss
+++ b/assets/sass/styles.scss
@@ -82,6 +82,20 @@ footer {
         min-height: auto;
         margin-bottom:0;
         background-color:#78C9E9;
+
+        a {
+            color: #fff;
+        }
+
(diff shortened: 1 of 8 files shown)


JorgeLu1s:
@roverwolf don't you need to add the `FOOTER_LEFT_TEXT` variable to the docker-compose* files?

roverwolf:
@JorgeLu1s When used, the `FOOTER_LEFT_TEXT` variable would have to be added the docker-compose.yml file, but it will probably not be used a lot so I had left it out.

I've now added it into the docker-compose files, as it does make it easier to maintain later.

PR #639, opened 2018-10-17
More Streamlined readme file

### LINKED ISSUE #638: Streamlined readme
As a user installing opensalt for the first time, I would like a readme that makes it more clear if/how/when to create usernames and passwords.

brandonopened:
- removed blockquotes to better highlight users to use the initial dev install script
- added note about virtual machine issues
- added headers if needed

closes #638

---
This change is [Reviewable](https://reviewable.io/reviews/opensalt/opensalt/639)

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,6 +14,8 @@ The code is intended to run using a set of docker containers using
 docker-compose so that it can be easily deployed in any Linux environment
 with docker installed.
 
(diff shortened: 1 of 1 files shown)


brandonopened:
Closing until we get a better idea of what 'should' be done here.

PR #644, opened 2018-10-24
Identifier display fix

### LINKED ISSUE #563: Change the way the URI is displayed and linked in the item detail panel

brandonopened:
Add identifier fields for CASE framework and Case Item. Closes #563 . *I don't see any tests to update with these two fields information but if I missed it I apologize.

---
This change is [Reviewable](https://reviewable.io/reviews/opensalt/opensalt/644)

### DIFF
--- a/assets/js/cftree/view-documentclass.js
+++ b/assets/js/cftree/view-documentclass.js
@@ -1357,7 +1357,7 @@ function ApxDocument(initializer) {
             let key, attributes, val;
             for (key in attributes = {
                 'officialSourceURL': 'Official URL',
-                'uri': 'CASE Framework URL',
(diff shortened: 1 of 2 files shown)


brandonopened:
@jmarks is the wording here correct? should it be CASE Document Identifier and CASE Item Identifier? (right now i just put Identifier for Document)

roverwolf:
Maybe just use `CASE Identifier` for both (and change the uris to `CASE URI` for both) ?

PR #648, opened 2018-10-26
moved CASE identifier to be firstabove the line

brandonopened:
patch to fully address #644 . 

Case URI And CASE identifier are now always the first two fields one sees in CfDoc or CfItem.

---
This change is [Reviewable](https://reviewable.io/reviews/opensalt/opensalt/648)


roverwolf:
Do we have to indicate here that CASE is a trademark due to IMS's trademark policy?  If so, should we just remove the word (leaving as simply URI and Identifier which should be enough)?

brandonopened:
Agree - made changes and updated test for framework page (item does not seem to need it - no test for identifier either?)

PR #656, opened 2018-11-01
check if association identifier was removed from the spreadsheet

### LINKED ISSUE #654: spreadsheet updater not updating associations on n levels
As an editor I want to use the update framework by spreadsheet to include N Levels for associations

JorgeLu1s:
Fixes #654

---
This change is [Reviewable](https://reviewable.io/reviews/opensalt/opensalt/656)

### DIFF
--- a/src/Service/ExcelImport.php
+++ b/src/Service/ExcelImport.php
@@ -74,7 +74,7 @@ class ExcelImport
                 $seq = null;
             }
 
-            if (array_key_exists($parentLevel, $smartLevels)) {
+            if (array_key_exists($parentLevel, $itemSmartLevels)) {
                 $smartLevels[$parentLevel]->addChild($item, null, $seq);
                 $children[$item->getIdentifier()] = $doc->getIdentifier();
             } else {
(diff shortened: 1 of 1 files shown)


src/Service/ExcelImport.php
```diff
@@ -96,15 +96,19 @@ public function importExcel(string $excelFilePath): LsDoc
 
         $sheet = $phpExcelObject->getSheetByName('CF Association');
         $lastRow = $sheet->getHighestRow();
+        $associationsIdentifiers = [];
+
         for ($i = 2; $i <= $lastRow; ++$i) {
             $association = $this->saveAssociation($sheet, $doc, $i, $items, $children);
+            $associationsIdentifiers[$this->getCellValueOrNull($sheet, 1, $i)] = null;
+
             if (null !== $association) {
```
roverwolf: This block, and the `$associations` array, does not appear to be used anymore and should probably be removed.

JorgeLu1s: you're right, I just removed it

PR #663, opened 2018-12-11
Adds custom fields for items

### LINKED ISSUE #504: addition of fields for OpenSALT for ACT internal use
Overview video: https://www.youtube.com/watch?v=VRtCunZe8c8  

**As an ACT internal user, I want to be able to edit fields not currently in CASE.** 

Lance (ACT)
- Can’t we do this in standard, reusable way that also anticipates CASE compliance?
- Notion of user-defined metadata fields assigned to an entity…for now in OpenSALT, this could be defined via DB insert skipping a UI for creation
- NOT hard coded field names
- Implement the minimum number of required field types
- String?, List?, Integer?
- Implement the minimum number of constraints by field type…Length, Precision
- Expose the fields (tab) as “Custom Fields” not “ACT Fields”
- (Eventually) extend CASE to support custom metadata

David: 

- There is a json column (named “extra”) in the database table that is useful for this purpose.
- Are the custom fields system wide or might they be framework specific?
- How should the fields be defined?
  - Internal name of the field -- should the name be namespaced for future sharing compatibility?
  - Is a display label needed for the field?  Or is the internal name enough?
  - Type of the field -- Just scalars?  Allow lists?  Enumerated or free-form?  Validation?
  - Should we use a standard/draft standard like JSON Schema to define the fields?

### LINKED ISSUE #645: custom fields
As a super editor, I want to be able to add custom fields to Cf Item entity via the extra column with a json object and have them show in the Cf Item screen.

roverwolf:
This is a separate PR that extends the addfields PR without the last commit from this morning.

Resolves #504 
Resolves #645 

---
This change is [Reviewable](https://reviewable.io/reviews/opensalt/opensalt/663)


src/Service/ExcelExport.php
```diff
@@ -227,14 +241,21 @@ protected function addItemRow(Worksheet $sheet, int $y, array $row): void
...
-            $columns = ['M', 'N', 'O'];
-            $c = 0;
-
-            foreach ($customFields as $field => $value) {
-                $sheet->setCellValue("$columns[$c]1", $field);
-                $sheet->setCellValue("$columns[$c]$y", $value);
-                ++$c;
+            $customFieldsData = $row['extra']['customFields'];
+
+            foreach ($customFieldsData as $field => $value) {
```
roverwolf: It would probably be better to loop over the columns 13 through 12+count(static::$customFields), and then get the value for the field for that column from the item and set it (i.e. just one loop instead of having an inner loop).

roverwolf: Refactored code pushed for your review.

tests/_support/Page/Framework.php
```diff
@@ -1414,4 +1414,21 @@ public function iCreateACustomField()
...
+     */
+    public function iDeleteACustomField()
+    {
+        $I = $this->I;
+
+        $I->amOnPage(self::$additionalFieldPath);
+        $I->see('test_additionalfield');
+
+        $I->click('Delete', Locator::combine('//table/tbody/tr', -1));
+        $I->wait(5);
```
roverwolf: Is the wait necessary here?  It is advised to avoid timed waits if there are other options (to allow the test to finish faster).

PR #685, opened 2019-02-06
Fix/missing associations

JorgeLu1s:
When importing the associations the validation was against the `$children` array; Since all the items were in that array, none association different than Is Child Of were getting imported.

---
This change is [Reviewable](https://reviewable.io/reviews/opensalt/opensalt/685)


roverwolf:
Does removing the `$children` check now still avoid the issue of duplicate children being created (if the isChildOf is listed as an association as well as being part of the smart level) that you originally added it for https://github.com/opensalt/opensalt/commit/0fb38f5062ffe2ae908c017a84703746ccc39088 ?

roverwolf:
@JorgeLu1s 
The _Scenario: 1013-1214 A Excel file can be uploaded and downloaded_ test fails due to:
```
We're sorry, we cannot load this document.  Please ensure this document is not already on the server, or see the Spreadsheet loading guide at docs.opensalt.org
```

I assume this means that we are uploading this framework more than once without either a) changing the UUIDs or b) deleting the previously uploaded framework.

Issue #667, opened 2018-12-18
Add Export to Association View


brandonopened:
Add export button to the associations view

http://frameworks-staging.act.org/cftree/doc/679/av

brandonopened:
While doing work related to this story, Jorge proposed showing all types by default and allowing for a filter drop-down on the association column using different logic. ("using the same logic as the regular search/filter".

PR #695, opened 2019-02-22
Remove current filters on the associations table

JorgeLu1s:
The current filter process was taking a lot of time to complete the task; they were removed and now the table is filtering with the native functions of the DataTable. Also this PR closes issue #667

---
This change is [Reviewable](https://reviewable.io/reviews/opensalt/opensalt/695)


roverwolf:
A few first things I've noted:

- The ability to delete an association seems to be removed by this change.
- The initial list shows reverse association types (such as *Is Parent Of*), but the type selector does not show those as options.  Might want to remove those from the list entirely (or add those types into the selector).
- The type selection box should probably be wider so that adding a few of them does not make it drop down to the next line.
- Hovering over a shortened statement no longer shows the full statement in a tool-tip popup.
- The `Excel` and `CSV` buttons are not formatted like other buttons in the UI, and there is no description as to what they do.

brandonopened:
I love the new way to select association types... but I don't see it working right now with association groups. And agree with Dave we should still be able to remove associations from this page etc.

PR #752, opened 2019-05-17
Replace "local:" in association URIs

roverwolf:
If the Origin/Destination object was not found do not leave "local:" in the URI, but instead replace with a generated absolute link.

---
This change is [Reviewable](https://reviewable.io/reviews/opensalt/opensalt/752)


roverwolf:
Resolves #753 

Also resolves #749 #750 (updated libraries)

Issue #758, opened 2019-05-24
Allow Association Groups to be created in Adopted Frameworks


brandonopened:
User Story: As an editor of a framework, I want to be able to add association groups to an adopted framework. 

Pursuant to a CASE spec discussion on Evolving Frameworks, one can add associations to adopted frameworks but currently I cannot add or edit association groups. Perhaps there is a reason for this structurally?

brandonopened:
for kicks I made a branch that would allow this - changed the toggle to be "editorrights" instead of "IsDraftStatus" which seems to satisfy potential use cases (ie allows editor in an organization to add and make associations even to an adopted framework)

Works without errors and the associations persist. however as stated wise to wait for more extensive use cases before this minor feature is submitted. 

<img width="883" alt="Screen Shot 2019-05-28 at 3 51 07 PM" src="https://user-images.githubusercontent.com/14943151/58517544-98af9c80-8160-11e9-8626-3dc34f4bd248.png">

jmarks:
Are you adding this to 2.2?

PR #795, opened 2019-06-11
Issue758 assoc groups visible

brandonopened:
Issue #758 

- made edit button dependent on "IsDraft" instead of editor rights
- made Association Groups show when EditorRights and not IsDeprecated (so it will show up for draft or adopted frameworks still only if that editor has Org etc rights)
- changed acceptance test

---
This change is [Reviewable](https://reviewable.io/reviews/opensalt/opensalt/795)

### DIFF
--- a/templates/framework/doc_tree/view.html.twig
+++ b/templates/framework/doc_tree/view.html.twig
@@ -205,11 +205,11 @@
                                             <button type="button" class="btn btn-default btn-sm" data-toggle="modal" data-target="#editDocModal" data-alt-document-disabled="true">
                                                 Edit
                                             </button>
-                                            {% if isDraft %}
+                                        {% endif %}
+                                        {% if editorRights and not isDeprecated %}
                                                 <button type="button" class="btn btn-default btn-sm" data-toggle="modal" data-target="#manageAssocGroupsModal" data-alt-document-disabled="true">
                                                     Manage Association Groups
                                                 </button>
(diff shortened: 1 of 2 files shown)


templates/framework/doc_tree/view.html.twig
```diff
@@ -201,15 +201,15 @@
                                             Export
                                         </button>
                                         {# use first version if we need to edit the adoptionStatus #}
-                                        {% if editorRights %}
+                                        {% if isDraft %}
```
roverwolf: One thing to note is that this will make it impossible to change the state after being changed to a non-draft state (ex. Adopted -> Deprecated).

templates/framework/doc_tree/view.html.twig
```diff
@@ -201,15 +201,15 @@
...
                                             <button type="button" class="btn btn-default btn-sm" data-toggle="modal" data-target="#editDocModal" data-alt-document-disabled="true">
                                                 Edit
                                             </button>
-                                            {% if isDraft %}
+                                        {% endif %}
+                                        {% if editorRights and not isDeprecated %}
```
roverwolf: Should probably also re-align the HTML inside this block.

templates/framework/doc_tree/view.html.twig
brandonopened: yeah i just noticed that, reworking...

